PR #2, opened 2015-08-03
fix typo

cl-lab-k:
found by aspell.

`Serverspec` is originated from `Server` + `RSpec`, not `Server` + `Spec`.
So 2nd `s` is not need to be capitalized.


philpep:
Merged, thanks !

PR #3, opened 2015-08-03
Docker Backend added

dmajere:
Good day, first of all, thanks for your work.
Here as backend for docker. For this backend the list of hosts are treated as docker images.
Usage:

```
testinfra --connection=docker --hosts=centos,alpine:3.1 test.py
```

I've been thinking about putting docker_host parameter in backend __init__ to command line parameters, but its occur that all the parameters are defined in plugin.py. You possibly should think about providing backend specific parameters in backend code, not in one central file.

### DIFF
--- a/testinfra/backend/docker.py
+++ b/testinfra/backend/docker.py
@@ -1,0 +1,59 @@
+# -*- coding: utf8 -*-
+# Copyright © 2015 Philippe Pepiot
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import unicode_literals
+from __future__ import absolute_import
+
+from testinfra.backend import base
+
+import os
+
+try:
+    from docker import Client
+except ImportError:
+    HAS_DOCKER = False
+else:
+    HAS_DOCKER = True
+
+
+class DockerBackend(base.BaseBackend):
+    _backend_type = "docker"
+
+    def __init__(self, host, *args, **kwargs):
+        if not HAS_DOCKER:
+            raise RuntimeError((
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,2 +1,3 @@
 pytest
 six
+docker-py
```
philpep: I would keep docker-py as an optional requirements, like paramiko and salt are.

See testinfra/backend/paramiko.py for a try/except import example

testinfra/backend/docker.py
```diff
@@ -0,0 +1,50 @@
...
+        self.cli.start(container)
+        inspect = self.cli.inspect_container(container)
+        if not inspect['State']['Running']:
+            raise RuntimeError("Docker Container Start Failed")
+
+        execution = self.cli.exec_create(container,
+                                         self.quote(command, *args),)
+        out = self.cli.exec_start(execution)
+        inspection = self.cli.exec_inspect(execution)
+        self.cli.remove_container(container, v=True, force=True, )
```
philpep: Related to #4 

I wonder if create/remove must occur for each command or for each test here.

Consider this test case:

````python
def test(Command):
    Command.check_output("echo foo > /bar")
    assert Command.check_output("cat /bar") == "foo"
```

With the current implementation, that test will fail.

philpep:
Hi thanks for your feedback !

Indeed, I consider a docker backend is a "must have" for testinfra. After thinking about the implementation I will be happy to merge it.

dmajere:
Moved docker-py to optional requirements as it's done for paramiko

testinfra/backend/docker.py
dmajere: Yep, for the first time i was doing contained creation in class init, but due to absence of class destructors in python and absence of setup/teardown phases for backend system i had to move create_container/remove_container to run method.

philpep:
Another issue is that docker-py doesn't seem to provide a way go get stdout and stderr separately

I opened https://github.com/docker/docker-py/issues/704

goodwillcoding:
I wonder if https://github.com/d11wtq/dockerpty can be help meanwhile.

dmajere:
Yes, docker-py is rather user-unfriendly. I'll look into dockerpty.

philpep:
Hey I've opened #13 with an implementation using the docker native client. Let me see if it works for you. Thanks

philpep:
Can one of the admins verify this patch?

philpep:
The docker backend was merged from #13 

Thanks for your contribution.

PR #5, opened 2015-08-03
Fix "testinfra --help" not showing usage

goodwillcoding:
Note: this is an ugly hack, because py.test currently does not allow
display of plugin usage

### DIFF
--- a/testinfra/plugin.py
+++ b/testinfra/plugin.py
@@ -1,4 +1,4 @@
-# -*- coding: utf8 -*-
+# -*- coding: utf-8 -*-
 # Copyright © 2015 Philippe Pepiot
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


philpep:
Hi, thanks for your contribution !

However I'm not gonna merge this because I just fixed it in 493dd0af5729aadb5007b2b021a6da34b040c11f

Now `testinfra --help` display py.test help (with testinfra specific options)

goodwillcoding:
Glad to help, thank you for testinfra.

To note, the difference in the fix I have is that it displays only help for the testinfra rather then all the py.test plugins which is a bit more confusing since it is so large. If that changes your mind about the PR, I am happy to update it to the easily merge with the existing code.

    usage: testinfra [-h] [--connection CONNECTION] [--hosts HOSTS]
                     [--ssh-config SSH_CONFIG] [--sudo] [--nagios]
    
    optional arguments:
      -h, --help            show this help message and exit
      --connection CONNECTION
                            Remote connection backend paramiko|ssh|safe_ssh|salt
      --hosts HOSTS         Hosts list (comma separated)
      --ssh-config SSH_CONFIG
                            SSH config file
      --sudo                Use sudo
      --nagios              Nagios plugin

philpep:
Yep py.test help is large and maybe confusing, but I prefer to keep a complete help because most of py.test options are useful with testinfra (-v, -k, --fixtures etc).

PR #9, opened 2015-08-07
Common backend interface

dmajere:
Good day,
I was thinking about common interface for all backends. This PR has an example of how it could be done, the main idea is to get dynamic backend loading without static lists in **init**.

### DIFF
--- a/testinfra/backend/__init__.py
+++ b/testinfra/backend/__init__.py
@@ -15,23 +15,22 @@
 
 from __future__ import unicode_literals
 
-from testinfra.backend import local
-from testinfra.backend import paramiko
-from testinfra.backend import salt
-from testinfra.backend import ssh
-
-BACKENDS = dict((klass.get_backend_type(), klass) for klass in (
-    local.LocalBackend,
-    ssh.SshBackend,
-    ssh.SafeSshBackend,
-    paramiko.ParamikoBakend,
-    salt.SaltBackend,
-))
+import importlib
 
 
 def get_backend(backend_type, *args, **kwargs):
+
+    backend_type_split = backend_type.split(".")
     try:
-        backend_class = BACKENDS[backend_type]
-    except KeyError:
+        if len(backend_type_split) == 1:
+            backend_module = importlib.import_module("testinfra.backend.%s" %
+                                                     backend_type)
+            backend_class = backend_module.Backend
+        else:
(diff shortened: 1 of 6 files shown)


testinfra/backend/ssh.py
```diff
@@ -21,14 +21,14 @@
 from testinfra.backend import local
 
 
-class SshBackend(local.LocalBackend):
+class Backend(local.LocalBackend):
```
philpep: I think `local.LocalBackend` should be changed to `local.Backend`

testinfra/backend/__init__.py
```diff
@@ -15,23 +15,13 @@
...
-))
+import importlib
 
 
 def get_backend(backend_type, *args, **kwargs):
     try:
-        backend_class = BACKENDS[backend_type]
-    except KeyError:
+        backend_module = importlib.import_module("testinfra.backend.%s" %
+                                                 backend_type)
```
philpep: I think the SafeSshBackend is broken here

philpep:
Hi, thanks for your PR. I also think it's a better way to do that.

And this is related to #10 that I've just opened, maybe you can implement this feature in this PR too ?

dmajere:
Hi, I'll try to do it during weekend.

dmajere:
Added some code to cover #10. Added a small test for get_backend

philpep:
Can one of the admins verify this patch?

cw-kuoka:
@dmajere @philpep Is anyone working on this?

Though the code looks very promising to me, it has been several months since this PR created.
Today it has some conflicts but I guess I have some time to resolve them and recreate PR.

mumoshu:
@dmajere @philpep Is anyone working on this?

Though the code looks very promising to me, it has been several months since this PR created.
Today it has some conflicts but I guess I have some time to resolve them and recreate PR.

philpep:
@mumoshu note that this can currently be done by overriding the ``TestinfraBackend`` fixture in a ``conftest.py`` and implementing a ``testinfra.backend.base.BaseBackend`` like class.

I'm not sure yet about implementing this feature like this PR does, I must be more convinced. What custom backend would you like to use ? Maybe I could try to implement it outside of testinfra, this will give me an idea of required changes (or not) in testinfra.

(sorry for the late response)

PR #13, opened 2015-08-07
Docker backend

philpep:
Docker backend using docker native client.

Usage: testinfra --connection=docker --hosts=debian:jessie

For each test function, testinfra will spawn a container, run function
on it, then kill and remove container.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,6 +11,7 @@ Vagrant.configure("2") do |config|
       vm.vm.provider "docker" do |docker|
         docker.build_dir = "images/#{vmname}"
         docker.has_ssh = true
+        docker.name = vmname
         if ['debian_jessie', 'centos_7', 'fedora_21'].include? vmname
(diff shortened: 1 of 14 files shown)


philpep:
Need to avoid `--connection=docker` without `--hosts`

philpep:
It seem this doesn't play well with parallel execution (using pytest-xdist) on a single host.

goodwillcoding:
I am guessing it would make sense to have the backend handle just running docker commands without start up and tear down at all (for example to test a running instance)

philpep:
Test this please

philpep:
retest this please

philpep:
@goodwillcoding you're absolutely right. Also, after thinking a bit, I think launching containers it out of scope of testinfra. There are already tools for that like vagrant, kitchen and it's easy to do it in a testinfra script (by using a hook or a fixture), allowing the user to configure many things like command, exposed ports, volumes etc.

PR #15, opened 2015-08-15
check if minion is up first before run_salt()

sh3llsh0ck:
Hello,

I have added a check before passing returning any commands from `run_salt()` using `test.ping` so if minion is down it will raise an error, i have tested it already but i will test it again soon to be sure enough.

Thanks

### DIFF
--- a/testinfra/backend/salt.py
+++ b/testinfra/backend/salt.py
@@ -53,4 +53,7 @@ class SaltBackend(base.BaseBackend):
             self, out['retcode'], out['stdout'], out['stderr'], command)
 
     def run_salt(self, func, args=None):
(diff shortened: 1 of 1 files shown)


philpep:
We can run a test.ping if cmd() returns an empty dict.
If test.ping returns "Minion not connected", raise an error "Minion not connected", otherwise raise an error "Unkown error".

Also if cmd() returns a dict like {"host": "Minion not connected"}, raise an error.

philpep:
Fixed in 6df82a96b4279de64c7d903590cb575a50579116

Just add "Minion not connected ?"  in the error message.

Thanks !

PR #16, opened 2015-08-17
get a list of all connected minions if asterisk passed to --hosts

sh3llsh0ck:
Hello,

As a SaltStack user , sometimes i need to use an asterisk to execute something on all minions, and thats also would be the case if i am testing a new state and would like to see how it works on all my minions, instead of typing the name of each minion (in my case i may test on 5 minions or more [different distors for example] ) it would be easier to pass an asterisk instead of typing the name of each minion. so i made the following modification:

```python
if (metafunc.config.option.hosts == "*" and
        metafunc.config.option.connection == "salt"):
           import salt.runner
           opts = salt.config.master_config("/etc/salt/master")
           runner = salt.runner.RunnerClient(opts)
           minions_list = runner.cmd("manage.up", [])
           metafunc.config.option.hosts = ",".join(minions_list)
```

it may be slow a little but its handy!

### DIFF
--- a/testinfra/plugin.py
+++ b/testinfra/plugin.py
@@ -105,8 +105,8 @@ def pytest_addoption(parser):
 
 def pytest_generate_tests(metafunc):
     if "_testinfra_host" in metafunc.fixturenames:
-        if (metafunc.config.option.hosts == "*" and 
-           metafunc.config.option.connection == "salt"):
+        if (metafunc.config.option.hosts == "*" and
+            metafunc.config.option.connection == "salt"):
             import salt.runner
             opts = salt.config.master_config("/etc/salt/master")
             runner = salt.runner.RunnerClient(opts)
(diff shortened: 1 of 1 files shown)


philpep:
Hi, I you can already do that with testinfra:

````python
import salt.client

def get_minions(pattern):
    client = salt.client.LocalClient()
    return client.cmd(pattern, "test.ping").keys()

testinfra_hosts = get_minions("*")

def test(Salt):
    assert Salt("test.ping")
````

This will run tests on all connected salt minions (without passing --hosts).

However, I think matching minions with salt "matchers" like you do can be a useful feature.

````shell
testinfra --hosts="web*"
testinfra --hosts="S@192.168.42.0/24"
etc
````

I'll consider this after thinking a bit about what that implies. But for now I cannot merge this as it.

I opened #17

Thanks

PR #18, opened 2015-08-21
Use vagrant and docker on travis

philpep:
http://blog.travis-ci.com/2015-08-19-using-docker-on-travis-ci/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,13 @@
+sudo: required
+services:
+  - docker
 language: python
 python: 2.7
-env:
-  - TOXENV=py27-travis
-  - TOXENV=py34-travis
-  - TOXENV=flake8
+before_install:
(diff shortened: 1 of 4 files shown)


daenney:
Why grab Vagrant at all? You can already do a docker pull and docker up through the CLI.

philpep:
@daenney I use vagrant to hide some complexity (ssh keys and config, docker run parameters etc)

````shell
$ vagrant up --provider=docker
$ vagrant ssh ubuntu_trusty
$ vagrant ssh-config > .vagrant/ssh-config
$ testinfra --ssh-config=.vagrant/ssh-config --hosts=ubuntu_trusty test_stuff.py
$ vagrant destroy -f
````

Doing the same thing, with the same ease, with only docker CLI would result to write a shell script equivalent to vagrant, but still less better than vagrant :)

PR #19, opened 2015-09-01
Properly check package.is_installed tests for rpm based system

mchugh19:
Noticed that package.is_installed didn't seem to properly verify that the package was actually installed. So here's a quick fix.


philpep:
Merged. Thanks !

PR #20, opened 2015-10-14
Add new Port module

brettlangdon:
Adding a new module `Port(address, port)` which can be used to test the state of active connections.

We wanted to be able to test if a service was running and listening on the appropriate ports. e.g.:

```python
test_nginx_listening(Port):
    http_port = Port("0.0.0.0", 80)
    assert http_port.is_listening

    https_port = Port("0.0.0.0", 443)
    assert https_port.is_listening
```

### DIFF
--- a/testinfra/modules/port.py
+++ b/testinfra/modules/port.py
@@ -1,0 +1,196 @@
+# -*- coding: utf8 -*-
+# Copyright © 2015 Philippe Pepiot
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import unicode_literals
+import re
+
+from testinfra.modules.base import Module
+
+
+class Port(Module):
+    """Test various active port attributes"""
+
+    def __init__(self, _backend, address, port):
+        self._address = address
+        self._port = port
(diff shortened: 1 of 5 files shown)


brettlangdon:
I updated here to fix the `flake8` issues. I have my local `flake8` line length to 120, so I didn't end up seeing them :(

Also fixed a few other things I saw that needed to be fixed.

testinfra/modules/port.py
```diff
@@ -0,0 +1,196 @@
...
+    def is_udp(self):
+        """Test if port is using udp
+
+        >>> Port("0.0.0.0", 53).is_udp
+        True
+        >>> Port("0.0.0.0", 22).is_tcp
+        False
+
+        """
+        return self.protocol and self.protocol.startswith("tcp")
```
philpep: should be "udp" here.

philpep:
Thanks for your contribution. Indeed I think this kind of module must be part of testinfra and I will be happy to merge it after thinking a bit more about the API.

- Extend to unix socket
- Naming: Port vs Socket or multiple modules like UnixSocket TCPSocket and UDPSocket
- list/count established connections / backlog
- Implementation using netstat, ss, /proc, conntrack, sockstat (bsd).... (I think you made the good choice with netstat, but I have to check if the output is the same on {Open,Net,Free}BSD)
- support port as name (/etc/services)

I don't have yet any fixed idea about this and will be happy to hear yours idea/solutions.

brettlangdon:
Yeah, after writing I toyed with the idea of naming it something more like `Connection` instead of `Port`, mostly was looking to get parity with `ServerSpec`, however, `Socket` I think works just as well.

How would you want to test a unix socket, wouldn't it be just a `File('/var/run/app.sock').is_socket`

philpep:
Hey, I've made a new PR with a WIP Socket module in #23 , tell me what to you think of the API on it

brettlangdon:
hey @philpep sorry for the delay. The other PR looks good, think moving to `Socket` is a good call.

philpep:
Closed in favour of #23

PR #22, opened 2015-10-15
Make File.mode be in octal form

brettlangdon:
**This is a backwards incompatible change**

Currently with `testinfra`, `File.mode` is provided as a decimal integer `644`, but it should be represented as an octal integer `0o644`, since this is how it is given to us by the os.

This change aligns with the Python standard library which uses an octal integer for `os.stat`.

```python
import os
import stat

s = os.stat('/etc/passwd')
mode = stat.S_IMODE(s.st_mode)
mode == 420
oct(420) == '0644'  # '0o644' in python 3
```

This also means we can utilize the file mode octal constants in the `stat` module for tests.

```python
import stat

def test_user_readable(File):
    passwd = File('/etc/passwd')
    assert passwd.mode & stat.S_IRUSR

def test_not_other_executable(File):
    passwd = File('/etc/passwd')
    assert not passwd.mode & stat.S_IXOTH

def test_file_mode(File):
    passwd = File('/etc/passwd')
    assert passwd.mode == (stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,7 +32,7 @@ Write your first tests file to `test_myinfra.py`::
         assert passwd.contains("root")
         assert passwd.user == "root"
         assert passwd.group == "root"
-        assert passwd.mode == 644
+        assert passwd.mode == 0o644
 
 
     def test_nginx_is_installed(Package):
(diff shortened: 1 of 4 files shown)


philpep:
Merged. Thanks !

brettlangdon:
Thanks for the quick merge here!

PR #23, opened 2015-10-19
New Socket module

philpep:
This is a WIP of a new Socket module.

TODO: BSD/Darwin

### DIFF
--- a/testinfra/modules/socket.py
+++ b/testinfra/modules/socket.py
@@ -1,0 +1,297 @@
+# -*- coding: utf8 -*-
+# Copyright © 2015 Philippe Pepiot
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


daenney:
Does this code actually check if something passed in as `unix://` is an actual socket? I don't see it doing so but it would be something along the lines of:

```python
import os, stat

def is_socket(path)
    return stat.S_ISSOCK(os.stat(path).st_mode))
```

philpep:
> Does this code actually check if something passed in as unix:// is an actual socket?

Nope. The beharvior will be (assuming `/sock` doesn't exists or isn't a socket neither a listening socket)

```python
>>> Socket("unix:///sock").is_listening
False
>>> Socket("unix:///sock").clients
[]
>>> # To test if /sock is a socket (warning doesn't test if the socket is listening)
>>> File("/sock").is_socket
````

File: testinfra/modules/socket.py [hunk 1]
```diff
@@ -0,0 +1,219 @@
...
+    elif ":" in address:
+        # tcp://127.0.0.1:22
+        # tcp://:::22
+        host, port = address.rsplit(":", 1)
+        port = int(port)
+    else:
+        # tcp://22
+        host = None
+        port = int(address)
+
```
daenney: This still allows for `Socket("tcp://0.0.0.0)`

philpep: Indeed. I've to check adress and port. I think this do the trick: https://stackoverflow.com/questions/319279/how-to-validate-ip-address-in-python/319298#319298

daenney: Yes but be careful not to use `aton` but `pton` so you also support IPv6.

Issue #21, opened 2015-10-14
Support OS X


brettlangdon:
Right now the test fail when running on OS X since `SystemInfo.type == "darwin"` is supported by the `File` module.

I resolved this in #20 by doing a check for `SystemInfo.type.endswith("bsd") or SystemInfo.type == "darwin"`. Hopefully it would be that easy to get OS X working in the other modules.

philpep:
Hi,

I don't have any OS X here. So I cannot add support for it. If you achieve this maybe http://docs.travis-ci.com/user/osx-ci-environment/ is a good test plaform for running OS X specific tests.

brettlangdon:
Oh, nice. Yeah, I am happy to dig into getting OS X working.

amitsaha:
I will try to pitch in with a fix here and there too.

jcftang:
OSX support would be nice, looking at the current travis config, it looks like that docker is used. If multiple OS's and distros are to be tested against then this is doc describes how to do it.

https://docs.travis-ci.com/user/multi-os/

gbirke:
I've created a pull request for the info in #177 
Tested on my local Mac (El Capital, Version 10.11.6)

Could not figure out how to test with travis, however

codylane:
@philpep - Hope you are doing well and thank you for creating this tool!  

I'm wondering if you are are still looking for volunteers and or help with creating MacOS integration(s)?   If so, i'm working on a a pretty significant addition to this library that will include the following:

1. Homebrew package support for OSX (MacOS) based hosts.
2. Enhancements and helpers to the `system_info` to help aide it OS identification and make testing less redundant.
3. A new backend VagrantBackend which will help facilitate the ease of automated CI and testing when using Vagrant.   You just tell the VagrantBackend hich Vagrant environment you want to use and the backend will handle the setup and teardown of your CI jobs via the backend letting you focus on creating new integrations like the docker host fixture that you have created in `test/conftest.py`
4. Travis CI enhancements to make sure OSX and base tests still work.

I'm still ironing out the problems with Travis CI but hope to have a pull request for you soon to address these enhancements.  

I'm looking for any constructive feedback to help make these additions better and look forward to hearing and working with you.

shakefu:
Is there any plans to add full support for "darwin" across the modules?

PR #26, opened 2015-10-21
Improve Mac OS X support (Issue #21)

amitsaha:
`readlink` behavior on Mac OS X is different than Linux and there is no
`sha256sum` command. Replace usage of these commands by Python interfaces.

Also assume, BSDFile for Mac OS X as well.

### DIFF
--- a/testinfra/modules/file.py
+++ b/testinfra/modules/file.py
@@ -16,6 +16,8 @@
 from __future__ import unicode_literals
 
 import datetime
+import hashlib
+import os.path
 
 from testinfra.modules.base import Module
 
@@ -69,7 +71,7 @@ class File(Module):
         >>> File("/var/lock").linked_to
         '/run/lock'
         """
-        return self.check_output("readlink -f %s", self.path)
+        return os.path.realpath(self.path)
 
     @property
     def user(self):
@@ -184,7 +186,7 @@ class File(Module):
             return GNUFile(_backend, None)
         elif SystemInfo.type == "netbsd":
             return NetBSDFile(_backend, None)
-        elif SystemInfo.type.endswith("bsd"):
+        elif SystemInfo.type.endswith("bsd") or SystemInfo.type == 'darwin':
             return BSDFile(_backend, None)
         else:
             raise NotImplementedError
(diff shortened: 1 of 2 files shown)


testinfra/modules/file.py
```diff
@@ -69,7 +71,7 @@ def linked_to(self):
         >>> File("/var/lock").linked_to
         '/run/lock'
         """
-        return self.check_output("readlink -f %s", self.path)
+        return os.path.realpath(self.path)
```
philpep: We can't do this for non local backend (paramiko, ssh, salt, docker, ...) neither for local backend with sudo.

amitsaha: @philpep I understand why for non-local backends. Why not with sudo?

philpep: I mead the `--sudo` option in testinfra

daenney:
Do any of these changes actually work with remote backends? I would expect we need to shell out for stuff like `sha256` too.

testinfra/modules/file.py
```diff
@@ -228,9 +230,7 @@ def md5sum(self):
 
     @property
     def sha256sum(self):
-        return self.check_output(
-            "sha256sum %s | cut -d ' ' -f 1", self.path)
-
```
daenney: On OS X this is simply `shasum -a 256 -p`

testinfra/modules/file.py
```diff
@@ -270,9 +270,7 @@ def md5sum(self):
 
     @property
     def sha256sum(self):
-        return self.check_output(
-            "sha256 < %s", self.path)
-
```
daenney: As noted above, `shasum -a 256 -p`

testinfra/modules/file.py
philpep: @daenney thanks ! But for linux I prefer keep sha256sum which is part of the GNU coreutils. So let's keep shasum only for OSX.

daenney: You'll need it for the BSD's too if I'm not mistaken. sha256 is gnu/coreutils which the BSD's don't ship with afaik.

PR #33, opened 2015-11-04
Service: Fix "is_enabled" for sysv services running under systemd

philpep:


testinfra/modules/service.py
```diff
@@ -65,47 +57,68 @@ def __repr__(self):
...
+            elif cmd.stdout.strip() == "disabled":
+                return False
+        elif self.service_backend == self.UPSTART:
+            if self.run(
+                "grep -q '^start on' /etc/init/%s.conf",
+                self.name,
+            ).rc == 0 and self.run(
+                "grep -q '^manual' /etc/init/%s.conf.override",
+            ).rc != 0:
+                return True
```
philpep: Need a specific test for '.override' file on ubuntu_trusty here.

PR #36, opened 2015-11-10
Ansible backend and module

philpep:


philpep:
Still need to fix the behavior when no ``--hosts`` is passed.

philpep:
I've implemented the host pattern matching too, allowing to write:

```testinfra --hosts='ansible://web*'```

Issue #43, opened 2015-12-17
'LinuxInterface' object is not callable


retr0h:
Having issues using `Interface`.

[jodewey:~/git … e-systems/deploy/metapod/vagrant_lite] CIA-232(+5/-5)+ 1 ± pip freeze | grep testinfra
testinfra==1.0.0.0a17

```
__________________________________________ test_has_internal_vip[ansible://mcp-rhel-7] ___________________________________________

Interface = <interface None>

    def test_has_internal_vip(Interface):
>       interface = Interface('eth1')
E       TypeError: 'LinuxInterface' object is not callable
```

PR #44, opened 2015-12-17
'LinuxInterface' object is not callable

retr0h:
The __call__ method was misnamed, preventing it from actually being
called.  Details in #43.


philpep:
Merged thanks ! (don't worry about the failed test, this is related to https://github.com/pytest-dev/pytest/issues/1259)

PR #45, opened 2015-12-20
adding the Process module

### LINKED ISSUE #24: Make A Process module
I'd be nice to have a Process module

  - process lookup (by pid, name, user etc)
  - user (integration with user module)
  - pid
  - cpu/mem usage
  - process arguments
  - sockets (integration with socket module)

hugoleodev:
With this issue I want  to resolve #24


testinfra/test/integration/test_centos_7.py
```diff
@@ -46,3 +46,13 @@ def test_systeminfo(SystemInfo):
     assert SystemInfo.release == "7"
     assert SystemInfo.distribution == "centos"
     assert SystemInfo.codename is None
+
+
+def test_process(Process):
+    sshd = Process("systemd")
```
daenney: `sshd = Process("systemd")`... that's a very unclear variable name.

hugoleodev:
@philpep The docker image philpep/testinfra:centos_7 is not finding the "which" command.

philpep:
Hi @hugoleodev I just pushed https://github.com/philpep/testinfra/commit/7f7aa2f0ebeaaf952f13e296d1f88ef9c1432817 to mitigate failed tests with pytest 2.8.5

Can you rebase your branch and remove the 'pytest==2.8.4' please ?

testinfra/modules/process.py
```diff
@@ -0,0 +1,48 @@
...
+    def __call__(self, name):
+        return self.__class__(self._backend, name)
+
+    @classmethod
+    def get_module(cls, _backend):
+        return Process(_backend, None)
+
+    def _get_column(self, column):
+        cmd = "ps -C {0} -o {1} --no-headers | head -1"
+        return self.check_output(cmd.format(self.process_name, column))
```
philpep: Don't use `format()` here. You need correct shell escaping handled by `check_output()`

```self.check_output("ps -C %s -o %s --no-headers | head -1", self.process_name, column)```

Comment (philpep) [inline testinfra/modules/process.py]: And escape the `%` (eg. `%%cpu`)

File: testinfra/modules/process.py [hunk 3]
```diff
@@ -0,0 +1,48 @@
...
+
+    def __call__(self, name):
+        return self.__class__(self._backend, name)
+
+    @classmethod
+    def get_module(cls, _backend):
+        return Process(_backend, None)
+
+    def _get_column(self, column):
+        cmd = "ps -C {0} -o {1} --no-headers | head -1"
```
philpep: I don't like the `head -1` here, multiple process can have the same name, so I was thinking about an api like:

```
>>> Process.filter(name="nginx")
[process1, process2, process3, process4]
>>> Process.get(name="nginx")
-> raise "MultipleProcessMatching" in case of multiple process, or only one process instance
```

`filter()` and `get()` will match process by name (contains) or exact match or regexp match or by pid, filter by user too etc.

philpep:
Indeed there is an issue with centos_7 image. A `$PATH` issue I suppose. Cannot reproduce here, only on travis :/

philpep:
Hi @hugoleodev , I've implemented the process module in #46 (still WIP) with the API I was thinking about. I'd like to have your feedback on it. Thanks !

hugoleodev:
Hi @philpep it was very awesome. I really like it. Good job. :+1:

philpep:
Closed in flavor of #46

muffl0n:
Thanks for this module, I find it really useful!

I was just wondering what your reasons are to parse the output of "ps" rather than accessing the information provided in /proc. I feel that using /proc (like here https://github.com/xolox/python-proc) is less error prone, but that's just a feeling. :)

philpep:
@muffl0n because testinfra is based on *remote* command execution, so handling /proc files would be painful :) (btw there is no /proc on BSDs)

PR #46, opened 2015-12-28
New process module

philpep:
Implementation for #24


philpep:
BSD support added (tested on Free/Net/Open BSD)

PR #66, opened 2016-02-21
Make Paramiko backend try to reconnect if connection became closed

AndrewPashkin:
For example, user might want host to be reverted to a pristine state
before each test run. In that case, the backend would fail, because
before this patch it was simply reused the old closed connection.

PR #71, opened 2016-02-29
Paramiko: Handle closed session

philpep:
When the connection is dropped (for instance if the target host is
restarted between tests), paramiko raises
"paramiko.ssh_exception.SSHException: SSH session not active". In this
case we try to reconnect (once).


AndrewPashkin:
I can confirm, that it fixes the issue addressed in #66

PR #74, opened 2016-03-12
Introduce LXD backend

AndrewPashkin:
It's still wip, I think tests should be updated too.

### DIFF
--- a/doc/source/backends.rst
+++ b/doc/source/backends.rst
@@ -79,3 +79,15 @@ The ansible backend use the `ansible python API
     $ testinfra --connection=ansible --hosts='web*'
 
 You can use an alternative `inventory` with the ``--ansible-inventory`` option
+
+LXD
+~~~~~~
+
+This backend provides capability of execution of commands in LXD_ containers.
+To run tests, using this backend, you can use such command::
+
+    $ testinfra --connection=lxd --hosts=container_name
+
+Where ``container_name`` is a name of running container.
+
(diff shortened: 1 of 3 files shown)


geoffwright:
@AndrewPashkin not sure why you closed this as it is a good addition. I was about to submit a PR doing exactly the same thing, but spotted this first.

Any chance this can be re-opened and merged?

AndrewPashkin:
@geoffwright, because I decided to open #76. Feel free to reopen.

geoffwright:
I saw #76 and think it also needs merging. Having both lxc and lxd backends would be useful.

Unfortunately I can't reopen.

AndrewPashkin:
@geoffwrigh, you can fork this my repo, pull my branch and make your own PR. Otherwise I don't know how to make it possible for you to take over this PR.

geoffwright:
I'll create a new PR from my fork and reference this one, I don't mind which one gets merged

philpep:
@AndrewPashkin @geoffwright can we handle both lxc and lxd with the same backend ?

For me there is still two issues:

- Implementation of `get_hosts()` (I'd take a look about providing a way to run local commands for this purpose)
- Tests ? Maybe installing an lxc instance inside the ubuntu docker image is possible ?

geoffwright:
I don't think you can combine them. From the container name you need to work out which tool created the container in order to use the right connection cmd. And there is the possibility that you have containers with the same name under lxc and lxd which leaves you guessing which container the user means.

PR #75, opened 2016-03-13
Add homebrew support to Package module

jcftang:
This just adds homebrew support for the package module.


ssbarnea:
Too old and not updated. Please reopen if still needed.

PR #77, opened 2016-03-21
Add simple supervisor service module

jcftang-r7:
This is a very naive module for checking if supervisor managed processes are running or not. It currently just executes on a host to check if something is running.

### DIFF
--- a/testinfra/modules/supervisor.py
+++ b/testinfra/modules/supervisor.py
@@ -19,8 +19,7 @@ from testinfra.modules.base import Module
 
 
 class Supervisor(Module):
-    """Test supervisor managed processes
-    """
+    """Test supervisor managed processes"""
 
     def __init__(self, name):
         self.name = name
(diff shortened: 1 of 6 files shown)


testinfra/modules/supervisor.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @classmethod
+    def get_module_class(cls, _backend):
+        Command = _backend.get_module("Command")
+        if (
+            Command.run_test("which supervisorctl").rc == 0
+            ):
+            return SupervisorService
+        else:
+            raise NotImplementedError
```
philpep: I think you doesn't need to implement this classmethod.

testinfra/modules/supervisor.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __repr__(self):
+        return "<service %s>" % (self.name,)
```
philpep: supervisor

testinfra/modules/supervisor.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class SupervisorService(Supervisor):
```
philpep: Cf my comment above, you can remove this class and move code to `Supervisor` class

testinfra/modules/supervisor.py
```diff
@@ -0,0 +1,60 @@
...
+    @property
+    def is_running(self):
+        # 0: program running
+        # anything else is not considered, the supervisorctl program
+        # always returns zero no matter what
+        return self.run_expect(
+                [0], "supervisorctl status %s | grep RUNNING" % self.name).rc == 0
```
philpep: I would implement methods to:

- List all services
- Get service status

After that `is_running` would become `self.status(name) == "RUNNING"`

philpep:
Hi, thanks.

Beside my comments, consider adding a test (install supervisor in debian_jessie docker image and put some dummy services inside the config).

jcftang-r7:
@philpep thanks for the comments, the PR was a first attempt at a supervisor module. I'll address the issues and submit more commits to this PR.

jcftang-r7:
@philpep I've addressed the issues above, all except the test part related to docker.

jcftang-r7:
I've added a dumb nc listener service and supervisord to the debian jessie docker image

testinfra/modules/supervisor.py
```diff
@@ -0,0 +1,57 @@
...
+        """Test if supervisord managed service is running"""
+        return self.status(self.name) == "RUNNING"
+
+    @property
+    def is_stopped(self):
+        """Test if supervisord managed service is stopped"""
+        return self.status(self.name) == "STOPPED"
+
+    @property
+    def is_enabled(self):
```
philpep: You can remove this method

testinfra/modules/supervisor.py
```diff
@@ -0,0 +1,57 @@
...
+        """Test if service is enabled"""
+        raise NotImplementedError
+
+    def list_services(self):
+        """Get services from supervisorctl and returns a dict"""
```
philpep: I would prefer `get_services` here. Also I could be better to return a list of dict with all informations here eg:

```
[
    {"name": "gunicorn", "pid": 42, "status": "RUNNING"},
    {"name": "celery", "pid": 4242, "status": "BACKOFF"},
]
```

testinfra/modules/supervisor.py
```diff
@@ -0,0 +1,57 @@
...
+        ret = {}
+        output = self.check_output("supervisorctl status")
+        for line in output.split("\n"):
```
philpep: You can use `splitlines()` instead of `split("\n")`

jcftang-r7:
@philpep I've made the suggested changes

philpep:
Merged in 3a5f4571141bd7a7c0bbd0a5c1b3a1b1d9df3b80 thanks !

PR #78, opened 2016-03-23
Return puppet facts when calling Facter

queeno:
When calling facter, make sure the puppet facts are included in the output.


philpep:
Merged. Thanks!

PR #80, opened 2016-03-30
Fix quoting bug which prevented running quote-requiring commands with sudo.

jaakkosipari:
Produced un-executable (even blocked the test run until ssh timeout) shell commands like this:

````sudo /bin/sh -c 'grep -qs -- '^first ' /tmp/testfile'````

### DIFF
--- a/testinfra/backend/base.py
+++ b/testinfra/backend/base.py
@@ -136,9 +136,10 @@ class BaseBackend(object):
             return command
 
     def get_command(self, command, *args):
+        command = self.quote(command, *args)
         if self.sudo:
             command = self.quote("sudo /bin/sh -c %s", command)
-        return self.quote(command, *args)
+        return command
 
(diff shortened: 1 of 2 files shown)


philpep:
Hi, thanks for the PR!

Could you add a test in test/integration/test_trusty.py ? I think a simple `assert Command.check_output("echo x | grep -qs '^x'") == "x"` should be enought (the trusty tests are run with and without sudo).

jsi-p:
I added a test case which mimics the real world situation where I first bumped into this problem. The test passes now, but I wasn't able to execute it manually with the "broken" code, so I'm not 100% that it tests the right thing.

philpep:
Fixup and merged in 25349b2 and v1.1.1 released.

I changed the test, the generated command was a valid shell command but not the good one ;)

Thanks !

Issue #84, opened 2016-04-04
Adding new modules


blarghmatey:
What is the level of interest for receiving pull requests to introduce new modules? For instance I have written a couple of modules for verifying Python and Ruby modules (e.g. `is_installed` and version checking), and I am wondering if it would be desirable to include these directly in TestInfra or if they would be better served in an extension package?

philpep:
Hi,

I'm open to modules that I can understand, test and maintain and that give the user a useful and unrestricted API (ie more than a oneliner that handle only a few cases).

I think handling python and ruby modules is a high interest for users of testinfra and I'll be happy to review your code (In this case, unrestricted API means virtualenvs handling ;) )

PR #86, opened 2016-04-08
Added Python and Ruby package modules

blarghmatey:
Added a module each for Python and Ruby package verification. Each
module exposes an `is_installed` and `version` property. They will each
accept an argument for the path to the preferred package manager binary
to use for determining which packages are installed.

### DIFF
--- a/testinfra/modules/__init__.py
+++ b/testinfra/modules/__init__.py
@@ -30,10 +30,13 @@ from testinfra.modules.socket import Socket
 from testinfra.modules.sysctl import Sysctl
 from testinfra.modules.systeminfo import SystemInfo
 from testinfra.modules.user import User
+from testinfra.modules.python_package import PythonPackage
+from testinfra.modules.ruby_package import RubyPackage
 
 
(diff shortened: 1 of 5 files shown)


philpep:
Hi thanks.

Like I said in #84 we must handle virtualenvs here. Also avoid mocking and make real tests in docker images (use debian_jessie).

blarghmatey:
Thanks, I'll update the tests as per your suggestion.

As far as virtualenv support, both the Python and Ruby package modules support a parameter to specify the path to the pip or gem binary respectively that will get used for determining which packages are installed. When you use a pip binary that is inside a virtualenv it will resolve the PYTHONPATH such that the packages listed will be the ones that are installed in the associated virtualenv.

georgealton:
I'd really like to see this

blarghmatey:
Closing pull request and opening a new one that has been rebased onto latest master.

blarghmatey:
Refer to https://github.com/philpep/testinfra/pull/99 for updated PR

PR #87, opened 2016-04-08
Fixed the `coding` setting on source files

blarghmatey:
The `coding` attribute was set to `utf8` which is not recognized as
being valid by Emacs, so I updated it to `utf-8` on all files.


philpep:
Hi, this was fixed in 6e7c8ad2b8fe4d005221943d743eac2fc4eadb9c

PR #92, opened 2016-04-12
Fix package version for packages on hold in dpkg

georgealton:
handle where package is on hold by using sed regexp

### DIFF
--- a/testinfra/modules/package.py
+++ b/testinfra/modules/package.py
@@ -74,7 +74,10 @@ class DebianPackage(Package):
 
     @property
     def is_installed(self):
-        return self.run_test("dpkg -s %s" % (self.name,)).rc == 0
+        out = self.check_output("dpkg-query -f '${Status}' -W %s" % (
+            self.name,)).split()
+        installed_status = ["ok", "installed"]
+        return out[0] in ["install", "hold"] and out[1:3] == installed_status
 
     @property
     def version(self):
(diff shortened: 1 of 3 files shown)


philpep:
Hi thanks.

I would add a test for this (in the debian_jessie image). Also I would be better to refactor a bit and avoid using `grep` here (which can hide error codes from dpkg) and post process directly in python

testinfra/modules/package.py
```diff
@@ -74,15 +74,14 @@ class DebianPackage(Package):
 
     @property
     def is_installed(self):
-        return self.run_test((
-            "dpkg-query -f '${Status}' -W %s | "
-            "grep -qE '^(install|hold) ok installed$'"), self.name).rc == 0
+        return self.run_test("dpkg -s %s" % (self.name,)).rc == 0
```
philpep: This will report uninstalled and non purged packages as installed. A test case is:

```
$ apt-get install redis-server
$ apt-get remove redis-server
$ dpkg -s redis-server >/dev/null
$ echo $?
0
```

philpep:
Looks almost perfect, thanks! Indeed removing the grep part here is better.

georgealton:
thanks phil, I tried a couple of packages with dpkg -s but didn't catch that, will commit an update now

philpep:
Merged. Thanks !

PR #98, opened 2016-04-29
MountPoint module

georgealton:
Module to allow testing mountpoint configuration


georgealton:
@philpep wondered if you had any comments/suggestions on this?

testinfra/modules/mountpoint.py
```diff
@@ -0,0 +1,120 @@
+# -*- coding: utf-8 -*-
+# Copyright © 2015 Philippe Pepiot
```
philpep: we are in 2016 and you can put your name here ;)

testinfra/modules/mountpoint.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @property
+    def filesystem(self):
+        """Returns the FileSystem type associated to a mount point
+
+        >>> MountPoint("/").filesystem
+        'ext4'
+
+        """
+        mountpoint = self.find_mountpoint()
```
philpep: Instead of calling `find_mountpoint()` on each property. Add a private attribute for caching values.

Something like:

```python
def __init__(self, path):
    self._attrs = None
    self.path = path
    super(MountPoint, self).__init__()

def _get_attrs(self):
    if self._attrs is None:
        self._attrs = self.get_mountpoint()
    return self._attrs

@property
def device(self):
   return self._get_attrs()["device"]
```

testinfra/test/test_modules.py
```diff
@@ -256,3 +256,11 @@ def test_ansible_module(TestinfraBackend, Ansible):
         assert variables["inventory_hostname"] == "debian_jessie"
         assert variables["ansible_user"] == "root"
         assert variables["group_names"] == ["ungrouped"]
+
+
+@all_images
+def test_mountpoint(MountPoint):
```
philpep: just testing on debian_jessie should be enough here (just remove the `@all_images` decorator)

testinfra/test/test_modules.py
```diff
@@ -256,3 +256,11 @@ def test_ansible_module(TestinfraBackend, Ansible):
...
+    root_mount = MountPoint('/')
+    assert root_mount.exists
+    assert 'rw' in root_mount.options
+    assert root_mount.filesystem == "aufs"
```
philpep: docker has other filesystem backend like overlay and zfs. Just checking the value is not empty should be enough.

philpep:
Hi, sorry to be late ;) Btw thanks for your work, it looks promissing.

Beside the comments I added, It would be nice to have a classmethod that return all available mount points (as list of MountPoint instances).

georgealton:
that's cool, thanks for the suggestions. I'll try and look at the list of MountPoint soon

georgealton:
had a go at a classmethod to generate the list of MountPoints

philpep:
Merged in 3586bed39b1fdcaa29b6ce9ac7d3b86965ce255a

Thanks !

PR #99, opened 2016-05-13
Add python and ruby package modules

blarghmatey:
Added a module each for Python and Ruby package verification. Each
module exposes an `is_installed` and `version` property. They will each
accept an argument for the path to the preferred package manager binary
to use for determining which packages are installed. This allows for use 
with virtualenvs or RVM gemsets


testinfra/modules/python_package.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
```
philpep: I would rather rename the PythonPackage module as Pip

testinfra/modules/python_package.py
```diff
@@ -0,0 +1,51 @@
...
+
+import re
+
+from testinfra.modules.base import Module
+
+
+class PythonPackage(Module):
+    """Test status and version of Python package."""
+
+    def __init__(self, pkg_name, pip_path='pip'):
```
philpep: I would prefer a virtualenv parameter here instead of pip_path (and exec pip from `os.path.join(virtualenv, "bin", "pip")`)

testinfra/modules/python_package.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        super(PythonPackage, self).__init__()
+        self.pip_dict = {}
+        self.pkg_name = pkg_name
+        cmd = '{0} list'.format(pip_path)
+        py_re = re.compile(r'^([A-Za-z-_1-9]+) \((.*?)(, .*)?\)')
+        for line in self.check_output(cmd).split('\n'):
+            if re.match(py_re, line):
+                name, version = re.search(py_re, line).groups()[:2]
+                self.pip_dict[name] = version
```
philpep: It's better to get package outside of `__init__` and also have a `get_packages()` classmethod that return the list of packages.

You can take inspiration from the MountPoint module https://github.com/philpep/testinfra/blob/master/testinfra/modules/mountpoint.py

testinfra/modules/ruby_package.py
```diff
@@ -0,0 +1,46 @@
...
+    @property
+    def is_installed(self):
+        return bool(self.gem_dict.get(self.pkg_name))
+
+    @property
+    def version(self):
+        return self.gem_dict.get(self.pkg_name)
+
+    def __repr__(self):
+        return "<ruby_package %s>" % (self.pkg_name,)
```
philpep: Same comment than the pip packages (rename as Gem, lazy get attributes, get_package method)

philpep:
I'm closing this PR. As we now have a PipModule. Feel free to implement a RvmModule. Thanks!

PR #100, opened 2016-05-14
Add check for crypted user password

tongpu:
This addition allows to check a crypted user password or a correctly disabled login for a system user.


philpep:
Hi, merged in 02a27a6a805e3192bd592db7d698b3b8d9827f6c with minor changes and support for BSDs. Thanks !

PR #102, opened 2016-05-19
User expires and system

elementalvoid:
Added two methods:

1. Check for user expiration.
2. Check if account is a system account.

Tox tests pass save the password test that was failing before the changes. All code passed flake8 checks.

### DIFF
--- a/testinfra/modules/user.py
+++ b/testinfra/modules/user.py
@@ -84,6 +84,16 @@ class User(Module):
         """Return the expiration days since epoch when account expires"""
         return self.check_output("getent shadow %s", self.name).split(":")[7]
 
+    @property
+    def system(self):
+        """Return True if system account, False otherwise"""
+        sys_min = self.check_output(
+            "awk '/SYS_UID_MIN/ {print $2}' /etc/login.defs")
+        sys_max = self.check_output(
+            "awk '/SYS_UID_MAX/ {print $2}' /etc/login.defs")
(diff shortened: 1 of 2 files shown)


elementalvoid:
Another push is pending. I'm adding tests for the new methods.

elementalvoid:
Test added for expires.

However, I'm having issues with the system attribute. Debian's min and max system uids are 100 and 999 respectively. The sshd user's uid is 105. The system attribute is returning False.

I'll fix this tomorrow and push another update.

philpep:
Hi, thanks !

The expiration look good. For the 'system' user I'd prefer a more reliable way (but not sure if it exists).

Also need to check implementation on BSDs (I can do it)

elementalvoid:
> For the 'system' user I'd prefer a more reliable way (but not sure if it exists).

I'm not sure there is one either. Maybe a better regex against the file or some failure checking. However, it's distribution dependent what the default values are and it's customizable. You literally have no idea what the system user ids might be when you land on some *nix box.

philpep:
Hi, I juste merged  aca298cf777de2d6d600c99f8dea6ee3af9520fb with the "account expiration" part (enhanced with BSD support, returning a datetime object and a real test).

I'll not keep the "system" account part because the way to get the information seems too weak for me (but can easily be done in a testinfra test).

I close this one. Thanks !

PR #123, opened 2016-08-31
Note that Socket module requires netstat

ashb:
It's available in most places, but it might not be available in docker containers so add a clarifying note to the docs -- save someone a bit of time if they get this error:

```
ipdb> Socket("tcp://80").is_listening
*** Failed: Unexpected exit code 127 for CommandResult(command=b'netstat -n -l -t', exit_status=127, stdout=b'', stderr=b'/bin/sh: 1: netstat: not found\n')
```

### DIFF
--- a/testinfra/modules/socket.py
+++ b/testinfra/modules/socket.py
@@ -69,6 +69,8 @@ class Socket(Module):
 
     ``socketspec`` must be specified as ``<protocol>://<host>:<port>``
 
+    This module requires the ``netstat`` command to on the target host.
+
     Example:
 
       - Unix sockets: ``unix:///var/run/docker.sock``


philpep:
Ok to add a note. But the error message is clear enough no ? "netstat: not found"

philpep:
Ok didn't see this was a PR. Thanks ! I restarted the build that has random failed due to debian mirrors issues.

testinfra/modules/socket.py
```diff
@@ -68,6 +68,9 @@ class Socket(Module):
     """Test listening tcp/udp and unix sockets
 
     ``socketspec`` must be specified as ``<protocol>://<host>:<port>``
+    
+    This module needs the  ``netstat`` binary to be installed in the
+    target environment otherwise it will not funciton.
```
philpep: s/funciton/function

But I'd reword like this "This module require the ``netstat`` command on target host".

ashb:
@philpep reworded and force pushed.

Issue #53, opened 2016-01-21
Ansible remote_user


emayssat-ms:
testinfra  -v --connection=ansible --ansible-inventory=inventories/ec2.py --hosts=key_Release001  main.py

The command above seems to work except 
1/ My entire inventory is dynamic
2/ SSH agent configuration is working and tested
3/ How can I specify the user to ssh on the remote box?
In my ansible setup, I do it with a directive in the playbook (remote_user). My local user is emayssat, my remote user is ubuntu.

Maybe should I use ansible.cfg? Other?

emayssat-ms:
http://docs.ansible.com/ansible/intro_dynamic_inventory.html#example-aws-ec2-external-inventory-script

https://aws.amazon.com/blogs/apn/getting-started-with-ansible-and-dynamic-amazon-ec2-inventory-management/ 

   just in case! ;-)

philpep:
Hi,

I see two solutions that you could try:

- have your inventory return `ansible_user` (ansible 2.X) or `ansible_ssh_user` (ansible 1.X).
- Add remote_user=ubuntu in your ansible.cfg

emayssat-ms:
I don't have control of the inventory script.
I tried the ansible.cfg solution and it works like a charm.
Thank you.
Emmanuel

PR #132, opened 2016-09-28
Added reference to ANSIBLE_REMOTE_USER

cheesesashimi:
I was struggling for a while to figure out how to set the username Ansible should use to connect. Rather than modifying the inventory / ansible.cfg as suggested in (here)[https://github.com/philpep/testinfra/issues/53], I thought it would make more sense to use one of Ansible's many [configurable environment variables](https://github.com/ansible/ansible/blob/devel/lib/ansible/constants.py).

I updated your backends doc to reflect that there is an environment variable one can use to control the remote user.

Any feedback / suggestions are gratefully accepted :).

### DIFF
--- a/doc/source/backends.rst
+++ b/doc/source/backends.rst
@@ -81,4 +81,6 @@ The ansible backend use the `ansible python API
     $ testinfra --connection=ansible --hosts=host1,host2
     $ testinfra --connection=ansible --hosts='web*'
 
-You can use an alternative `inventory` with the ``--ansible-inventory`` option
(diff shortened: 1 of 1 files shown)


philpep:
Hi,

Testinfra use ansible API, so all environment variables and configurations that affect ansible behavior applies to testinfra as well. I'd be better to explain this in the docs and add reference to relevant ansible docs.

Thanks !

cheesesashimi:
You're right, that makes more sense.

PR #138, opened 2016-10-18
expand path when using --ssh-config

karmab:
necesary to make the following work properly 
testinfra --connection=ssh --sudo --ssh-config=~/.ssh/config


philpep:
Thanks !

PR #139, opened 2016-10-21
Fix parallel execution with Salt backend

BenoitKnecht:
Running tests in parallel with xdist used to fail with the Salt backend
because each worker would get a list of tests in a different order.

This commit ensures that the Salt backend returns hosts in a
deterministic order.


philpep:
Merged, thanks !

PR #145, opened 2016-11-08
Check for presence of commands with command -v

rosstimson:
I encountered an issue when trying to use the new [Amazon Linux container](https://docs.aws.amazon.com/AmazonECR/latest/userguide/amazon_linux_container_image.html) image as it does not come with the `which` command. 

The `which` command isn't always available, especially on more minimal Docker images.  A more POSIX compliant way of checking for the presence of commands is with `command -v`.

### DIFF
--- a/testinfra/modules/base.py
+++ b/testinfra/modules/base.py
@@ -35,7 +35,7 @@ class Module(object):
         return out
 
     def run_test(self, command, *args, **kwargs):
-        """Run command and check it return an exit status of 0 or 1
+        """Run command and check it return an exit status of 0, 1 or 127
 
         :raises: AssertionError
         """
--- a/testinfra/modules/base.py
+++ b/testinfra/modules/base.py
@@ -39,7 +39,7 @@ class Module(object):
 
         :raises: AssertionError
         """
-        return self.run_expect([0, 1], command, *args, **kwargs)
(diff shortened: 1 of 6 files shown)


rosstimson:
Ack, my bad, apologies for not checking for tests.

I hope I've updated with what you were looking for.

testinfra/modules/base.py
```diff
@@ -35,11 +35,11 @@ def run_expect(self, expected, command, *args, **kwargs):
         return out
 
     def run_test(self, command, *args, **kwargs):
-        """Run command and check it return an exit status of 0 or 1
+        """Run command and check it return an exit status of 0, 1 or 127
 
         :raises: AssertionError
         """
-        return self.run_expect([0, 1], command, *args, **kwargs)
+        return self.run_expect([0, 1, 127], command, *args, **kwargs)
```
philpep: don't modify the behavior or `run_test()`, directly use `run_expect([0, 1, 127], "command -v ...")` where you need it instead.

philpep: @rosstimson would you be ok that I fix and squash your commit keeping your authoring ?

rosstimson: @philpep Yes that would be great.  Apologies, it has been a bad week at work so I've not had time to revisit this.

philpep:
closed in flavor of #152

PR #147, opened 2016-11-10
New PipPackage module

### LINKED ISSUE #117: Add PipPackage module
Would be nice to have something like `Package` module but for `pip` :-)

philpep:
Check installed and outdated packages using pip.
Virtualenv support.

Closes #117

### DIFF
--- a/doc/source/modules.rst
+++ b/doc/source/modules.rst
@@ -80,6 +80,12 @@ Package
    :members:
 
 
+PipPackage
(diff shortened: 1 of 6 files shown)


wissemmahjoub:
Hi,
i'm using this module, to verify if ansible is outdated or not , in debian9 .
but i have a wrong result .. 
it tell me ansible is outdate, but is not ..
why ?

PR #149, opened 2016-11-12
Ignore decoding errors when falling back to backend encoding

### LINKED ISSUE #148: Local connection decoding errors
```
pip (9.0.1)
pkg-resources (0.0.0)
pluggy (0.4.0)
py (1.4.31)
pytest (3.0.4)
setuptools (28.8.0)
six (1.10.0)
tox (2.4.1)
virtualenv (15.0.3)
wheel (0.30.0a0)
```

Simple reproduction:

```
import testinfra
>>> testinfra.get_backend('local://').get_module('Command').check_output('echo \xaa')
  File "<stdin>", line 1, in <module>
  File "testinfra/modules/base.py", line 51, in check_output
    out = self.run(command, *args, **kwargs)
  File "testinfra/modules/base.py", line 23, in run
    return self._backend.run(command, *args, **kwargs)
  File "testinfra/backend/local.py", line 33, in run
    return self.run_local(self.get_command(command, *args))
  File "testinfra/backend/base.py", line 167, in run_local
    command = self.encode(command)
  File "testinfra/backend/base.py", line 214, in encode
    return data.encode("ascii")
UnicodeDecodeError: 'ascii' codec can't decode byte 0xaa in position 5: ordinal not in range(128)
>>> 
```

uSpike:
Fixed #148

### DIFF
--- a/testinfra/backend/base.py
+++ b/testinfra/backend/base.py
@@ -207,7 +207,7 @@ class BaseBackend(object):
         try:
             return data.decode("ascii")
         except UnicodeDecodeError:
-            return data.decode(self.encoding)
+            return data.decode(self.encoding, errors='ignore')
 
     def encode(self, data):
         try:


philpep:
Close this since I think it's a misconfiguration of the test environment and we should not hide such errors.

uSpike:
By misconfiguration, do you mean the default encoding if the local connection?  Since the local connection defaults to `ascii` encoding, this code block effectively becomes:

```
         try:
             return data.decode("ascii")
         except UnicodeDecodeError:
             return data.decode("ascii")
```
That doesn't make any sense to me.

philpep:
@uSpike  The local connection encoding is determined by ``python -c 'import locale;print(locale.getpreferredencoding())``, is it "ascii" for you ?

uSpike:
@philpep I'm testing this inside a LXC container.

```
 $ lxc exec test -- /bin/bash -c 'python -c "import locale;print(locale.getpreferredencoding())"'
ANSI_X3.4-1968
 $ lxc exec test -- /bin/bash -c 'export LANG="en_US.UTF-8"; python -c "import locale;print(locale.getpreferredencoding())"'
UTF-8
```

So, the issue is the default encoding/locale of the platform. Do you think that there should be a check for when the connection encoding is `ascii`?

```
         try:
             return data.decode("ascii")
         except UnicodeDecodeError:
             if self.encoding == 'ascii':
                 return data.decode(self.encoding, errors='ignore')
             else:
                 return data.decode(self.encoding)

PR #154, opened 2016-12-01
adding a release property to package module

rfgtid:
Hi

*Rationale*: I have found testinfra very useful for my work. But in my case, the relase info returned by RPM package manager is relevant for us.

This Pull Request adds an 'release' property to the Package module for package managers that support this property. Otherwise returns an `NotImplementedError` exception. This should not break any existing tests already written.

### DIFF
--- a/testinfra/modules/package.py
+++ b/testinfra/modules/package.py
@@ -41,6 +41,15 @@ class Package(Module):
         raise NotImplementedError
 
     @property
+    def release(self):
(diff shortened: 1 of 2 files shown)


rfgtid:
Thanks, I assume that no more changes on my side are pending. 
I'm not yet used to the community contribution and CI processes awith travis, tox, dockers, etc. :-).

PR #160, opened 2016-12-11
Bump ssh fedora release

jesselang:
Tests were failing, and I'm trying to get a PR in, so...


jesselang:
Meh, I'll split this out.

PR #164, opened 2016-12-15
fixed trailing LineFeed bug

rfgtid:
I have detected that SystemInfo.type string got a hidden trailing CR.
A strip() gets away with this.


philpep:
This occur with what OS and connection backend ?

I think this is the purpose of `check_output()` (return stdout without trailing newline).

You can replace this code https://github.com/philpep/testinfra/blob/master/testinfra/modules/base.py#L55 by `return out.stdout.rstrip()` and I'll be happy to merge it.

Thanks !

philpep:
EDIT: ``rstrip("\r\n")`` should be more appropriate

rfgtid:
Sorry, RHEL6 using ssh backend

PR #165, opened 2016-12-16
Fix for RpmPackage

mr-ice:
parsing rpm -qi was inefficient and ineffective. 

rpm -qi and line.split(":", 1)[1].strip()  were giving

```
>>> Package('acl').version
2.2.49                            Vendor: Red Hat, Inc.
```

While technically this starts with the same characters as the version, it has a bunch more we don't want to see and that have no business in the version string.

 rpm -q --queryformat gives direct output of the values we want.

Package('acl').version == '2.2.49'

This test now works:
```
def test_package_acl(Package):
    acl = Package("acl")
    assert acl.is_installed
    assert acl.version == '2.2.49'
    assert acl.release == '6.el6'

```

### DIFF
--- a/testinfra/modules/package.py
+++ b/testinfra/modules/package.py
@@ -135,17 +135,6 @@ class OpenBSDPackage(Package):
 
 class RpmPackage(Package):
 
-    def _get_package_info(self, header):
-        # Name        : bash
-        # Version     : 4.2.46
-        # Release     : 0.g9b0f0e9
-        # ...
-        out = self.check_output("rpm -qi %s", self.name)
-        for line in out.splitlines():
-            if line.startswith(header):
-                return line.split(":", 1)[1].strip()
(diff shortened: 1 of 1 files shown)


mr-ice:
Linking to Issue #166

mr-ice:
Acknowledge tox failure, I'll try to get to the bottom of that (whitespace?).

mr-ice:
Having trouble with tox because it invokes pip differently from how I do within my virtualenv.  I'm behind a proxy and I have to set up SSL certificates for it to work.  When I run pip manually it checks ~/.pip/pip.conf for the location of certs.   tox is not so conscientious.

mr-ice:
Update incoming.

mr-ice:
I don't have any preference on releaseversion(), it is inferior to having each separately queryable, so I have dropped it.

_get_package_info() is also dropped.

philpep:
Merged. Thanks !

Issue #151, opened 2016-11-19
RFC enhanced readability of testinfra fixtures


philpep:
This is a proposal of a future syntax to replace current fixture system:

```python
# capitalized fixture name that must be repeated for each test declaration
def test_something(Command, Service, Sudo, File, Package):
    [...]
```

Proposition 1: Just one fixture to access all modules and remove capitalization of first letter (could be named Server or Testinfra or Connection or Backend)

```python
def test_something(Server):
    assert Server.service("nginx").is_running
    assert Server.package("nginx").version = "1.4.3"
```

Proposition 2: Import

```python
import pytest
from testinfra.modules import Service

# on whole module or for each testinfra test
pytestmark = pytest.mark.testinfra

def test_something():
    assert Service("nginx").is_running
```

Migration path: DeprecationWarning for v1, drop old usage in v2

tinproject:
I really like the proposition 1, having one only entry point could low the entry barrier for testinfra (only one thing to remember) and I think it's better for auto-complete tools to show what are all the available possibilities.

But, the name `Server` could be misleading because you also could test containers, I would use something more meaningful and explicit to name that 'superfixture', like: `testinfra_subject`, or `testinfra_sut` (sut == subject under test), or something like that, perhaps more succinct. 

Also the capitalization breaks PEP8, its a nuisance that the linter is always yelling about that function arguments must be lowercase, and also punch the principle of least astonishment:
- In the current assert case `assert Service("nginx").is_running`, you think that you are invoking a class, obtaining an instance for the nginx service and then assert a property of that instance, that's ok.
- On the proposition 1 case: `assert Server.service("nginx").is_running`, what do you could think? that you are calling an staticmethod ot a classmethod of the server class? and then, what do you obtain? a Server instance or a Service instance?... that's confusing.

To increase the readability the first I would do is to drop the capitalized names of fixtures.

tinproject:
Another advantage of proposition 1 is that have a whole server fixture could enable to end up testing more than one server in a single file. It has to be a way to define different fixtures for different servers, but this approach goes more in line with the philosophy of pytest and fixture/test composition.

philpep:
Thank you for your advice @tinproject 

I prefer proposition 1 too. About the name of the fixture, what about just ``testinfra`` ? This will clash with a potential ``import testinfra`` statement but it's explicit and clear that the fixture expose testinfra features.

[...]

PR #170, opened 2016-12-24
single fixture 'host'

philpep:
Related to #151

I opted to name the fixture ``host`` and use un-camel-case module name (eg. SystemInfo -> system_info).

Currently the fixture should work and old test suite should pass

TODO:

- [x] Find a proper way to share cache betwen host and backend.
- [x] Deprecate old fixtures
- [x] Migrate tests to use new fixture
- [x] Document the new fixtures
- [x] make shortcut (eg ``host.check_output(cmd)`` instead of ``host.command.check_output(cmd)``

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,22 +27,22 @@ Install testinfra using pip::
 
 Write your first tests file to `test_myinfra.py`::
 
-    def test_passwd_file(File):
-        passwd = File("/etc/passwd")
+    def test_passwd_file(host):
+        passwd = host.file("/etc/passwd")
         assert passwd.contains("root")
         assert passwd.user == "root"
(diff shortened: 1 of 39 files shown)


philpep:
I consider the feature as "ready", I plan to release in a few days if no blocking issues are found before.

Please test using ``pip install 'git+https://github.com/philpep/testinfra@single-fixture#egg=testinfra``

From the user POV, this should just add pytest warnings like:

```
WC1 None File fixture is deprecated. Use host fixture and get File module with host.file
```

CC @retr0h

retr0h:
Looks like a nice cleanup.  Users of Molecule will need to change their tests, but I can introduce that at a minor release.

PR #180, opened 2017-01-26
Restore backend command if sudo fails

manics:
If a Sudo Command fails the backend command isn't restored by the context manager. If you then run another Command in a different test method the Sudo context is still active, which may lead to it failing if that user was unprivileged.

This PR adds a test demonstrating the problem, and a fix.

Travis is failing, but I think that's for reasons unrelated to this PR- it's also [failing on the master branch](https://travis-ci.org/manics/testinfra/builds/195503413).


philpep:
Merged and released in 1.5.3

Thanks !

PR #182, opened 2017-02-08
Update example code for testing docker images

swordsp:
File.content returns a byte string, so the test will fail in Python 3.


philpep:
Merged, thanks !

PR #186, opened 2017-02-14
Service: fix Upstart is_running test

BenoitKnecht:
When a service is managed by Upstart, the "service my-service status"
command returns exit code 0 even when the service is stopped, which
means UpstartService.is_running was true even when the service wasn't
running.

So instead of relying on the exit code, match the "running" string in
the output of "status my-service".

### DIFF
--- a/testinfra/modules/service.py
+++ b/testinfra/modules/service.py
@@ -128,6 +128,14 @@ class UpstartService(SysvService):
             # Fallback on SysV
             return super(UpstartService, self).is_enabled
 
+    @property
+    def is_running(self):
+        cmd = self.run_test('status %s', self.name)
+        if cmd.rc == 0 and len(cmd.stdout.split()) > 1:
+            return 'running' in cmd.stdout.split()[1]
+        else:
+            return super(UpstartService, self).is_running
+
 
 class FreeBSDService(Service):
 


philpep:
Hi, thanks for contributing !

I'll be great to have some test of this, ubuntu_trusty docker image use upstart (not started as PID 1 though). Or maybe point to relevant issue/documentation or a real example that I can try to add tests for.

BenoitKnecht:
I thought about tests, but my understanding is that Upstart isn't running *at all* in the `ubuntu_trusty` docker image, right?

One possibility would be using the [ubuntu-upstart](https://hub.docker.com/_/ubuntu-upstart/) image; it's deprecated, but it should work just fine for the purpose of testing.

I tried the following:

```console
# docker run --rm -p 2222:22 -d ubuntu-upstart
69d287ce80578e6fde4edac5eb286423a7c5c08946864b4357a06a9cb8e87f51
# ssh -p 2222 root@localhost
The authenticity of host '[localhost]:2222 ([::1]:2222)' can't be established.
ECDSA key fingerprint is SHA256:XrQkoy7LBVMTSkpSz+08lv/TRWyplQ4D/Y6jJYagLeE.
Are you sure you want to continue connecting (yes/no)? yes
Warning: Permanently added '[localhost]:2222' (ECDSA) to the list of known hosts.
root@localhost's password: 
Welcome to Ubuntu 14.04 LTS (GNU/Linux 3.19.0-43-generic x86_64)

 * Documentation:  https://help.ubuntu.com/

The programs included with the Ubuntu system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.

Ubuntu comes with ABSOLUTELY NO WARRANTY, to the extent permitted by
applicable law.

root@69d287ce8057:~# status ssh; echo $?
ssh start/running, process 42
0
root@69d287ce8057:~# status cron; echo $?
cron stop/waiting
0
root@69d287ce8057:~# status urandom; echo $?
status: Unknown job: urandom
1
root@69d287ce8057:~# service urandom status; echo $?
4
```

It shows that `ssh` and `cron` are both managed by Upstart (with `ssh` running and `cron` stopped), while `urandom` is not managed by Upstart and is stopped.

Let me know if you'd like me to implement tests based on this image, or if you would prefer adapting `ubuntu_trusty` to allow for Upstart to be running.

philpep:
Ok thanks, this looks clearer to me now and I'm fine with your patch. Since upstart is deprecated, let's not add too much efforts on this.

For the record: "status foobar" exit 0 with no output when init is not upstart, that's why we fallback on SysV in this case.

Merged, thanks !

PR #190, opened 2017-03-15
SocketHost module that takes hostname as variable

shaidar:
Socket module takes IP address as host variable, however in some of our tests, we need to pass in a hostname as opposed to an IP. SocketHost module has two functions based on [serverspec](http://serverspec.org):
- is_resolvable
  * check if provided hostname can be resolved to an IP address 
- is_reachable
  * check if hostname provided is reachable by opening a socket connection and then closing it. If hostname resolves to multiple IP addresses, only the first returned IP will be contacted.


shaidar:
Expecting pylint to fail given the similarity between the Socket and SocketHost module

testinfra/modules/sockethost.py
```diff
@@ -0,0 +1,120 @@
...
+    if port is not None:
+        try:
+            port = int(port)
+        except ValueError:
+            raise RuntimeError("Cannot validate port '%s'" % (port,))
+
+    return protocol, host, port
+
+
+class SocketHost(Module):
```
philpep: I'm not sure it's worth to add a new module for this. Maybe adding ``is_resolvable()`` and ``is_reachable()`` as classmethod to the existing ``Socket`` module could be better

testinfra/modules/sockethost.py
```diff
@@ -0,0 +1,120 @@
...
+        if socketspec is not None:
+            self.protocol, self.host, self.port = parse_socketspec(socketspec)
+        else:
+            self.protocol = self.host = self.port = None
+
+    @property
+    def _attrs(self):
+        self._attrs_cache = {}
+        if not self._attrs_cache:
+            t = socket.gethostbyname_ex(self.host)
```
philpep: We can't use python socket modules here. Testinfra is a remote execution testing framework, so it should work through ssh. Use shell commands instead.

testinfra/modules/sockethost.py
```diff
@@ -0,0 +1,120 @@
...
+    @property
+    def is_reachable(self):
+        """Test if host is reachable
+
+        Assumption is that if protocol provided is not specified as udp,
+        then we are going to use socket.SOCK_STREAM.
+
+        """
+        if self.protocol == 'udp':
+            sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
```
philpep: Same apply here, use remote commands instead of python socket module.

philpep:
I'm closing this one as we now have a `socket` module. Thanks!

PR #193, opened 2017-04-11
modules: interface: return None if speed not defined

BenoitKnecht:
On Linux, some interfaces don't have a speed attribute, such as `lo` for
instance. In that case, `cat /sys/class/net/lo/speed` returns:

  cat: /sys/class/net/lo/speed: Invalid argument

which causes Interface.speed to throw an exception.

This commit makes Interface.speed return None in those cases, instead of
throwing an AssertionError.

### DIFF
--- a/testinfra/modules/interface.py
+++ b/testinfra/modules/interface.py
@@ -62,8 +62,9 @@ class LinuxInterface(Interface):
 
     @property
     def speed(self):
-        return int(self.check_output(
-            "cat /sys/class/net/%s/speed", self.name))
+        speed = self.run("cat /sys/class/net/%s/speed", self.name)
+        if speed.rc == 0:
+            return int(speed.stdout)
 
     @property
     def addresses(self):


philpep:
Hi, thanks for contributing !

If the interface name does not exists it will return None as well and I think it's better to raise an error in this case.

```
% cat /sys/class/net/doesnotexist/speed
cat: /sys/class/net/doesnotexist/speed: No such file or directory
% echo $?
1
```

Maybe we should keep the current behavior and assume we can't test speed of such interfaces.

BenoitKnecht:
You're right, I didn't think about that. It still seems counter-intuitive to me to throw an assertion error when the speed is undefined though.

How about checking whether the interface exists or not during `__init__()`, since the interface properties (`speed`, `address`) don't make sense if it doesn't; the exception could be thrown there.

Otherwise, I would do `assert self.exists` at the beginning of `speed`, and then use the same logic as proposed in this PR; but it seems like we'd end up having that assertion on every single property, so having it in `__init__()` would be more straightforward.

Let me know what you think.

tyll:
It seems to me that an except in `__init__()` if the interface does not exist breaks workflows that want to assert that an interface does not exist, i.e. `assert not interface.exists()`. Also one might use the same interface object to assert that an interface does not exist, create it externally and then assert that does exists now or the other way around.

PR #199, opened 2017-04-26
Add new module -- Network

pichuang:
Implement be_reachable by command `ping` and be_resolvable by command `host`

Signed-off-by: Phil Huang <phil_huang@edge-core.com>

### DIFF
--- a/doc/source/modules.rst
+++ b/doc/source/modules.rst
@@ -91,6 +91,10 @@ host
 
        :class:`testinfra.modules.user.User` class
 
+    .. attribute:: network
+
+       :class:`testinfra.modules.network.Network` class
+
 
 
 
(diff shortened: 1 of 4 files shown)


pichuang:
It seems that I loss `coverage.py` data file
https://travis-ci.org/philpep/testinfra/jobs/225917041#L2920-L2923

skhoroshavin:
Any reason why this isn't merged?

philpep:
Hi, thanks for contributing and sorry for the long delay I took to review this MR.
Several people want this kind of module, but I think we should write down some specs prior providing implementations (because API is frozen after a release). Also I would like to provide a valuable module for this, with advanced features which make having a module worth. I suggest you to discuss the spec on https://github.com/philpep/testinfra/issues/358

Thanks!

PR #212, opened 2017-06-21
Add hacking to test requirements

bartoszcisek:
Without this module test would not pass


philpep:
We already have hacking in ``dev-requirements.txt`` and it's not required for all tests (only ``tox -e flake8``). I don't like too much hacking both ``dev-requirements.txt`` and ``test-requirements.txt``, this is confusing.

I suggest to only keep test-requirements.txt and declare explicitly requirements for each environment in tox.ini

PR #219, opened 2017-07-05
adding checking if service is enabled for OpenBSD

dmizelle:
Realized that we're not able to see if services are enabled on OpenBSD. With OpenBSD >= 5.8, `rcctl` now has an `ls` command where you can get output like:

```
$ rcctl ls on
check_quotas
cron
ntpd
pf
pflogd
postfix
sensuclient
sndiod
sshd
syslogd
```

### DIFF
--- a/testinfra/modules/service.py
+++ b/testinfra/modules/service.py
@@ -173,14 +173,13 @@ class OpenBSDService(Service):
 
     @property
     def is_enabled(self):
-        # rcctl ls is only available on 5.8 or newer
-        if self.check_output('uname -r') >= '5.8':
-            for service in self.check_output("rcctl ls on").splitlines():
-                if service and service == self.name:
-                    return True
+        if self.name in self.check_output('rcctl ls on').splitlines():
+            return True
+        if self.name in self.check_output('rcctl ls off').splitlines():
             return False
-        else:
-            return NotImplementedError
+        raise RuntimeError(
+            "Unable to determine state of {0}. Does this service exist?"
+            .format(self.name))
 
 
 class NetBSDService(Service):
--- a/testinfra/modules/service.py
+++ b/testinfra/modules/service.py
@@ -24,7 +24,7 @@ class Service(Module):
     - Linux: detect Systemd or Upstart, fallback to SysV
     - FreeBSD: service(1)
     - OpenBSD: ``/etc/rc.d/$name check`` for ``is_running``
-      (``is_enabled`` is not yet implemented)
+      ``rcctl ls on`` for ``is_enabled`` (only OpenBSD >= 5.8)
     - NetBSD: ``/etc/rc.d/$name onestatus`` for ``is_running``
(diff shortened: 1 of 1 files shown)


testinfra/modules/service.py
```diff
@@ -173,7 +173,14 @@ def is_running(self):
...
     def is_enabled(self):
-        raise NotImplementedError
+        # rcctl ls is only available on 5.8 or newer
+        if self.check_output('uname -r') >= '5.8':
+            for service in self.check_output("rcctl ls on").splitlines():
+                if service and service == self.name:
+                    return True
+            return False
+        else:
+            return NotImplementedError
```
philpep: Thanks for adding this ! I tested and it work well !

I'd remove the check on OpenBSD version, I think it's ok to let the command fail on older versions.
Also the loop seems useless, you can write ``return self.name in self.check_output("rcctl ls on").splitlines()``

An other issue with this implementation is that a service that not exists will always been reported as 'not is_enabled'. For instance this could occur by a typo in the service name, like ssh instead of sshd. I'd expect an error here.

Something like:

```python
@property
def is_enabled(self):
    if self.name in self.check_output('rcctl ls on').splitlines():
        return True
    if self.name in self.check_output('rcctl ls off').splitlines():
        return False
    raise RuntimeError('could not determine if {0} is enabled, does this service exist ?'.format(self.name))
```

dmizelle: Ah! I didn't even think about what if the service doesn't even exist.

I'll push a new commit soon. Thanks for the feedback!

dmizelle:
And here's out the output looks when a service isn't defined:

```
=================================== FAILURES ===================================
__________ test_openbsd_services[ansible://openbsd60.ci.dev] __________
[gw2] linux2 -- Python 2.7.12 /home/devon/git/ansible/venv/bin/python2

host = <testinfra.host.Host object at 0x7fa49181cfd0>

    def test_openbsd_services(host):
        if host.system_info.type == 'openbsd':
            assert host.service('sshd').is_enabled
            assert not host.service('unbound').is_enabled
>           assert host.service('foobar').is_enabled

tests/test_base.py:11:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <service foobar>

    @property
    def is_enabled(self):
        if self.name in self.check_output('rcctl ls on').splitlines():
            return True
        if self.name in self.check_output('rcctl ls off').splitlines():
            return False
        raise RuntimeError(
            "Unable to determine state of {0}. Does this service exist?"
>           .format(self.name))
E       RuntimeError: Unable to determine state of foobar. Does this service exist?

../../venv/local/lib/python2.7/site-packages/testinfra/modules/service.py:182: RuntimeError
```

philpep:
Merged, thanks !

PR #218, opened 2017-07-04
Add Vagrantbackend and OSX Support

codylane:
Hello -

Thanks again for creating such an awesome framework, I hope that these new additions are acceptable and will be merged into master.

Summary Of Changes
================
1. New VagrantBackend to more easily work with Vagrant environments.  The original plan for this backend was to be used by testinfra and spin up MacOS vagrant boxes (setup and teardown) when vagrant tests are specified using the 'vagrant_sut' pytest fixture.
2.  A few new enhancements to `system_info` to reduce redundant operating system testing.
3.  A new homebrew package implementation for OSX based hosts.
4.  A new LaunchD service implementation for OSX based hosts.
5.  Flake8 and Pylint fixes to all of the framework.
6.  Small refactors in some major components of testinfra to make readability a little easier IMHO.

Things I Still Need To Do
==================
I plan on tackling these in the next day or two but wanted to get this pull request in ASAP.

- [ ] I still need to add some tests for LaunchDService.
- [ ] I still need to add tests for system_info enhancements.

 I'm happy to refactor, explain or fix any constructive suggestions you may have so that this will be merged into master.  This has been something I've been working on for the past few months and would love some feedback.

PR #226, opened 2017-07-29
Initial implementation for OSX support + system_info enhancements

codylane:
Hi @philpep,

I went ahead and closed my original pull request #218 per your recommendations in favor to open this one.  

I know this is still pretty large, but I don't know how to make it any smaller due to the ways testinfra is implemented.  

I first started out adding new features for the `system_info` module only to find out I had to add other Darwin specific modules just to get `system_info` to work and for the existing tests to pass.  I know it's still a lot to go through but I really do hope this can get merged into your awesome code base.

Anyway, I'm looking forward to working with you on this and I hope this can be merged into master.  Please let me know if you have any questions or concerns.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,13 +15,7 @@ matrix:
       sudo: required
       language: python
       python: 2.7
-      env: TOXENV=flake8
-    - os: linux
-      dist: trusty
-      sudo: required
-      language: python
-      python: 2.7
-      env: TOXENV=pylint
+      env: TOXENV=flake8,pylint
     - os: linux
       dist: trusty
       sudo: required
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,41 @@
-sudo: required
+---
 services:
   - docker
-env:
-  matrix:
-    - TOXENV=py27,ansible1,py34
-    - TOXENV=flake8
-    - TOXENV=pylint
-    - TOXENV=sphinxdoc
-language: python
-python: 2.7
+
+matrix:
+  include:
+    - os: linux
+      dist: trusty
+      sudo: required
+      language: python
+      python: 2.7
+      env: TOXENV=py27,ansible1,py34
+    - os: linux
+      dist: trusty
+      sudo: required
+      language: python
+      python: 2.7
+      env: TOXENV=flake8
+    - os: linux
+      dist: trusty
+      sudo: required
+      language: python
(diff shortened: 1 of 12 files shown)


codylane:
@philpep - Just wondering if you've had a chance to review this request or if I'm able to answer any questions?

philpep:
@codylane sorry for my late reply, I was on vacations.

I cannot understand this pull request, see the unified diff on "files changed", there is a lot of changes. Please cleanup the commits and split changes in individual pull requests (1 feature = 1 PR).

codylane:
Hi @philpep -

Thanks for your response and I really hope you enjoyed your holidays.

While I can totally appreciate your review process I find it a bit odd that each commit has to be a separate pull request especially when adding something like OSX support.  

I really tried to keep the commit volume down for a single feature but in practice I found that these requirements may need to be relaxed when there are other factors like breaking your entire test suite.   This is exactly what happened when I introduced the system_info module changes.  

Also, some of those commits are not even mine and came in due to the fact that my branch was behind upstream master and this pull request got those additions as as side effect.  Perhaps my git foo is bad or there is some standard that I'm not aware of when it comes to open source contributions.   

As mentioned, if I put each commit as a PR as (1:1) then you will have broken tests because of the way tests are implemented in testinfra.  

In order to add OSX support I had to update the Travis CI configuration which is also a dependency for passing tests.  To address your concern, I think you'll find that most of my commits are mostly tests.  But, I can totally appreciate your review process I just don't want to spend months and months going back and forth.

This pull request is "one feature" which extends the 'system_info' module that also has accompanying tests.  But after running the tests with just the 'system_info' additions the test facility fails due to dependencies with other modules like "MountPoint" and "Socket" modules when you are running the tests on an OSX based system.  To me, it is unacceptable for a contributor to submit code to a upstream project that would introduce broken tests.  So I kindly ask that you please revise your process. 

I hope that you will reconsider reviewing these commits and consider this addition.    I've already spent way too much time on this so I don't see any benefit to break out each commit into a PR.  I put a lot of thought into this to make sure it wouldn't harm testinfra in a bad way.

codylane:
Hi @philpep - I know you are busy but I'm just wondering what your thoughts and recommendations are to either proceed or close this pull request?  I support your decision either way.

I'd prefer not to create my own version of testinfra just to add OSX support.  ;)

PR #230, opened 2017-08-19
allow specifying a config dir for salt module

gtmanfred:
This is specifically useful in kitchen-salt for me, since we are putting the configs in /tmp/kitchen/etc/salt

Thanks,
Daniel

### DIFF
--- a/testinfra/modules/salt.py
+++ b/testinfra/modules/salt.py
@@ -42,13 +42,15 @@ class Salt(InstanceModule):
         if self._host.backend.HAS_RUN_SALT:
             return self._host.backend.run_salt(function, args)
         else:
+            cmd_args = []
             cmd = "salt-call --out=json"
             if local:
                 cmd += " --local"
             if config is not None:
-                cmd += " -c {0}".format(config)
+                cmd += " -c %s"
+                cmd_args.append(config)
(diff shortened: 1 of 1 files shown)


testinfra/modules/salt.py
```diff
@@ -45,6 +45,8 @@ def __call__(self, function, args=None, local=False):
             cmd = "salt-call --out=json"
             if local:
                 cmd += " --local"
+            if config is not None:
+                cmd += " -c {0}".format(config)
```
philpep: Hello, thanks for contributing !

Looks good to me except that if "config" need to be shell escaped (if it contains spaces or something else)

I'll do:

```python
cmd_args = [function]
[...]
if config is not None:
    cmd += " -c %s"
    cmd_args.append(config)
cmd += " %s" + len(args) * " %s"
cmd_args += args
```

Let me know if you have time to fix this, if not I'll be happy to modify your commit and to merge it. Thanks !

gtmanfred: if we put the function first in cmd_args, it will be passed to `-c` instead, but I can do the string escape stuff.

gtmanfred:
Changes made.

philpep:
Merged, thanks !

PR #231, opened 2017-08-24
add support for pywinrm

gtmanfred:


testinfra/backend/winrm.py
```diff
@@ -0,0 +1,66 @@
...
+    """Run command through winrm command"""
+    NAME = "winrm"
+
+    def __init__(self, hostspec, password=None, ssl=True, verify_ssl=True,
+                 transport=None, *args, **kwargs):
+        self.host, self.username, self.port = self.parse_hostspec(hostspec)
+        self.password = password
+        self.ssl = ssl
+        self.verify_ssl = verify_ssl
+        self.transport = transport or 'ntlm'
```
philpep: I'd use ``__init__(..., transport='ntlm') and self.transport = transport`` here.

testinfra/backend/winrm.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def run(self, command, *args, **kwargs):
+        return self.run_winrm(self.get_command(command, *args))
+
+    def run_winrm(self, command, *args):
+        conn_args = {
+            'endpoint': '{proto}://{host}:{port}/wsman'.format(
+                proto='https' if self.ssl else 'http',
+                host=self.host,
+                port=self.port
```
philpep: What if we don't supply a port in ``hostspec`` ? It will be set to ``None`` and probably create a wrong ``endpoint`` here.

I'd suggest

```python
'endpoint': '{proto}://{host}{port}/wsman'.format(
    [...]
    port=':{}'.format(self.port) if port else ''
)
```

gtmanfred: looks good, will do

testinfra/backend/winrm.py
gtmanfred: ahh yeah, strings are immutable, so that makes sense, just habit with passing dictionaries around probably

gtmanfred:
also updated here.

Thanks!
Daniel

philpep:
Perfect ! Merged, thanks !

gtmanfred:
Awesome, thanks.

I am switching my tests to run off your git repo.  Any ideas on when there might be a release with this stuff in it?

Thanks!
Daniel

philpep:
I'll be happy to make a new release with this. It will be awesome if you add a bit of docs and an invocation example in https://github.com/philpep/testinfra/blob/master/doc/source/backends.rst :)

gtmanfred:
Can do!

PR #235, opened 2017-08-30
socket: handle AF_INET6 as AF_INET4

### LINKED ISSUE #234: Socket module doesn't return ipv4 client connected to tcp6 socket
Hello,

tcp6 socket actually works for both ipv4 and ipv6. See [section 3.7 - Compatibility with IPv4 Nodes of RFC 3493 - Basic Socket Interface Extensions for IPv6](https://tools.ietf.org/html/rfc3493#section-3.7).

Thus, when a daemon only creates a tcp6 socket, it can be used by ipv4 client. For example, the gerrit server only creates a tcp6 socket, and ipv4 clients looks like this in netstat:

```bash
$ netstat -nea | grep 29418
tcp        0      0 192.168.0.109:53594     192.168.0.109:29418     ESTABLISHED 994        44153592  
tcp6       0      0 :::29418                :::*                    LISTEN      995        44154634
```

This behavior makes it impossible to look for gerrit clients with testinfra Socket module:
* host.socket("tcp://:::29418") works but  .clients returns empty list
* host.socket("tcp://0.0.0.0:29418") or "tcp://29418" doesn't works

It seems like "tcp://$port" should works when there is any tcp6 socket.

TristanCacqueray:
IPv6 listening socket are actually IPv4 compatible socket.

Closes #234

### DIFF
--- a/testinfra/modules/socket.py
+++ b/testinfra/modules/socket.py
@@ -113,7 +113,7 @@ class Socket(Module):
         allipv4 = (self.protocol, "0.0.0.0", self.port) in sockets
         allipv6 = (self.protocol, "::", self.port) in sockets
         return (
-            all([allipv4, allipv6])
+            any([allipv6, all([allipv4, allipv6])])
             or (
                 self.host is not None
                 and (


philpep:
Merged, thanks !

PR #238, opened 2017-08-31
Add docs for winrm and salt

gtmanfred:
Also add flags to the options for the winrm backend can be passed from the command line

Somehow I forgot to push my changes to my repo on the last one :/ odd

Thanks!
Daniel


gtmanfred:
What is going on? Why are these getting closed?

https://github.com/philpep/testinfra/compare/master...gtmanfred:master

PR #240, opened 2017-09-01
Use a modified pipes.quote for winrm backend

gtmanfred:
The winrm backend fails if the strings in it are quoted with single
quotes instead of double quotes.

@philpep I ran into problems using this on linux, probably because of the way that linux shells handle single quotes vs double quotes.

But if this gets in, it should be good for a release.

Thanks!
Daniel

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -7,6 +7,7 @@ disable=
     bad-continuation,
     fixme,
     locally-disabled,
+    wrong-import-order
 max-branches=13
 max-bool-expr=6
 max-args=8
--- a/test/test_backends.py
+++ b/test/test_backends.py
@@ -16,6 +16,7 @@ import pytest
 import testinfra.backend
 from testinfra.backend.base import BaseBackend
 from testinfra.backend.base import HostSpec
+from testinfra.backend.winrm import _quote
 BACKENDS = ("ssh", "safe-ssh", "docker", "paramiko", "ansible")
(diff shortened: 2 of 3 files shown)


gtmanfred:
Well that is fun.

pylint thinks it should be 

```
import string
import six
```

and flake8 thinks it should be
```
import six
import string
```

philpep:
I'd trust flake8 (hacking) for that, you can disable ``wrong-import-order`` in ``pylintrc``

testinfra/utils/__init__.py
```diff
@@ -14,11 +14,42 @@
...
+            if c not in _safechars:
+                break
+        else:
+            if not s:
+                return "''"
+            return s
+
+    # use single quotes, and put single quotes into double quotes
+    # the string $'b is then quoted as '$'"'"'b'
+    return '"' + s.replace('"', '"\'"\'"') + '"'
```
philpep: Can you put this code in winrm.py since it's used only for this backend ?

Also it will be great to add relevant tests for this.

gtmanfred: added.

gtmanfred:
When you get a chance to get back to this, after it's merged if we could get a release, I would greatly appreciate it.

Thanks!
Daniel

gtmanfred:
Thanks for the release!
Daniel

PR #246, opened 2017-09-11
Fix documentation typo

achaussier:


philpep:
Good catch ! Merged, thanks !

PR #251, opened 2017-09-26
replace the png logo file by an svg one

douardda:
The font is nto exactly the same as the one in the PNG file, but hey.


philpep:
Merged, thanks !

PR #256, opened 2017-10-05
[docs] Update --nagios documentation to remove tracebacks

arthurlogilab:
I don't often see very verbose outputs to nrpe checks, so I would recommend using --tb none for that. (The operator can then go and run the test manually to get more information)


philpep:
Hi Arthur, thanks for contributing ! :)

It's seems to be `no` and not `none`. Also I wonder if `--tb line` could be more appropriate ?

marco-m:
hello @arthurlogilab any news?

marco-m:
merci! Now it is @philpep turn :-)

philpep:
Merged, thanks Arthur!

PR #263, opened 2017-10-29
Fix for 'linked_to' test for MacOS

elliotweiser:
While running testinfra against a Mac OS system, I noticed that
the `linked_to` test uses `readlink -f` to evaluate a canonicalized
real path by recursively following all symlinks. Though this behavior
works with GNU's `readlink`, it does not work the same way for BSD's
version. In fact, it errors out because BSD's `readlink` does not
understand the `-f` flag.

It is possible to install the GNU equivalent for `readlink` on MacOS
via Homebrew. However, it seems like poor design to force this method
to have a dependency on Homebrew.

I found a StackOverflow post showing how to recreate the behavior of
GNU's `readlink` using BSD's `readlink`. This change is based on the
solution provided in that post, which fixes the behavior of the
`linked_to` test for systems using the BSD version of `readlink`
(e.g. MacOS). It should behave just like `readlink -f` on Linux systems.

Reference: https://stackoverflow.com/questions/1055671/how-can-i-get-the-behavior-of-gnus-readlink-f-on-a-mac

### DIFF
--- a/testinfra/modules/file.py
+++ b/testinfra/modules/file.py
@@ -267,6 +267,24 @@ class BSDFile(File):
         return self.check_output(
             "sha256 < %s", self.path)
 
+    @property
+    def linked_to(self):
+        link_script = '''
+        TARGET_FILE='{0}'
+        cd `dirname $TARGET_FILE`
+        TARGET_FILE=`basename $TARGET_FILE`
+        while [ -L "$TARGET_FILE" ]
+        do
+            TARGET_FILE=`readlink $TARGET_FILE`
+            cd `dirname $TARGET_FILE`
+            TARGET_FILE=`basename $TARGET_FILE`
+        done
+        PHYS_DIR=`pwd -P`
+        RESULT=$PHYS_DIR/$TARGET_FILE
+        echo $RESULT
+        '''.format(self.path)
+        return self.check_output(link_script)
+
 
 class NetBSDFile(BSDFile):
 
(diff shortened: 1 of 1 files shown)


elliotweiser:
@philpep Let me know if there's anything I can do to improve this change (or if you have a better fix in mind).

philpep:
Hi, sorry for the late response and thanks for contributing!

I think FreeBSD, OpenBSD and NetBSD have `readlink -f` working as expected
https://www.freebsd.org/cgi/man.cgi?query=readlink
https://man.openbsd.org/readlink.1
https://www.unix.com/man-page/netbsd/1/readlink/

```
     When invoked as readlink, only the	target of the symbolic link is
     printed.  If the given argument is	not a symbolic link and	the -f option
     is	not specified, readlink	will print nothing and exit with an error.  If
     the -f option is specified, the output is canonicalized by	following
     every symlink in every component of the given path	recursively.  readlink
     will resolve both absolute	and relative paths, and	return the absolute
     pathname corresponding to file.  In this case, the	argument does not need
     to	be a symbolic link.
```

So your issue concern only MacOS, can you create a new DarwinFile(BSDFile) that implement your version of `linked_to()` ? Thanks!

elliotweiser:
No worries and thanks for the feedback! I'll take care of it when I grab some spare cycles.

elliotweiser:
@philpep All fixed! Thanks again for looking at this!

philpep:
Merged, thanks!

PR #265, opened 2017-11-15
Feature: Add is_valid method to systemd.

Poogles:
After deploying some new service overrides I realised I'd made a spelling mistake in one of the fields, this should catch these by asserting there's no warnings.

I've tested these manually on a linux box, should be simple for you to do, just add a junk field to the unit file and it should now fail.

Open to name changes on `is_valid`, it seemed most in keeping with the package names.  Unsure where to add docs as well?

### DIFF
--- a/testinfra/modules/service.py
+++ b/testinfra/modules/service.py
@@ -125,9 +125,8 @@ class SystemdService(SysvService):
         # stdout for anything.  Nothing means no warns/errors.
         # Docs at https://www.freedesktop.org/software/systemd/man/systemd
         # -analyze.html#Examples%20for%20verify
-        if not cmd.stdout and not cmd.stderr:
-            return True
-        return False
+        assert (cmd.stdout, cmd.stderr) == ("", "")
+        return True
 
 
 class UpstartService(SysvService):
--- a/testinfra/modules/service.py
+++ b/testinfra/modules/service.py
@@ -113,6 +113,22 @@ class SystemdService(SysvService):
         # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=760616
         return super(SystemdService, self).is_enabled
 
+    @property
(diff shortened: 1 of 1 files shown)


testinfra/modules/service.py
```diff
@@ -113,6 +113,22 @@ def is_enabled(self):
...
+        else:
+            name = self.name + ".service"
+        cmd = self.run("systemd-analyze verify %s", name)
+        # A bad unit file still returns a rc of 0, so check the
+        # stdout for anything.  Nothing means no warns/errors.
+        # Docs at https://www.freedesktop.org/software/systemd/man/systemd
+        # -analyze.html#Examples%20for%20verify
+        if not cmd.stdout and not cmd.stderr:
+            return True
+        return False
```
philpep: Hi, thanks for contributing! And sorry for my late review.

This looks useful to testinfra and I learned how to verify a systemd service, thanks! :)
Nevertheless I'll do a small modification to have a better reporting because I guess users will use it with `assert service.is_valid` and we could provide more explanation by having stdout and stderr in the assertion reporting:

Maybe ```assert (cmd.stdout, cmd.stderr) == ("", "")``` ?

Poogles: No problem.

I thought of this when I was writing it but for some reason didn't do it.  Updated the PR.  Can rebase if you wish.

philpep:
Merged, thanks!

Documentation could be added by using a docstring, but I think it will not work because the doc document the "Service" module and not the SystemdService. Maybe you could implement a `is_valid` o n Service that ``raise NotImplementedError()` and with a docstring explaining how this can be used and a note that say that is only available for Systemd based systems.

Poogles:
Awesome cheers.  Do you know when you're going to have a new release with this in?

PR #270, opened 2017-11-23
iptables module

scootersmk:
Addresses one of the few features I currently use in serverspec that testinfra lacks. Open to suggestions on code. Thoughts?

### DIFF
--- a/testinfra/modules/iptables.py
+++ b/testinfra/modules/iptables.py
@@ -1,0 +1,50 @@
+# coding: utf-8
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import unicode_literals
+
+from testinfra.modules.base import Module
+
+
+class Iptables(Module):
+    """Test iptables rule exists"""
+
+    def __init__(self, rule, table=None):
+        self.rule = rule
+        self.table = table
+        super(Iptables, self).__init__()
+
+    @property
+    def exists(self):
+        """Test if file exists
+
+        >>> host.file("/etc/passwd").exists
+        True
+        >>> host.file("/nonexistent").exists
+        False
+
+        """
+        rules = self.get_rules()
+        return self.rule in rules
+
+    def get_rules(self):
+        rules = []
+        cmd = "iptables -S"
+
+        if self.table:
+            cmd += " -t %s" % (self.table)
+
+        for line in self.check_output(cmd).splitlines():
+            line = line.replace("\t", " ")
+            rules.append(line)
+        return rules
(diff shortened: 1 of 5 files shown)


testinfra/modules/iptables.py
```diff
@@ -0,0 +1,50 @@
...
+    @property
+    def exists(self):
+        """Test if file exists
+
+        >>> host.file("/etc/passwd").exists
+        True
+        >>> host.file("/nonexistent").exists
+        False
+
+        """
```
philpep: docstring should be updated here.

Also the autodoc could be added in doc/source/modules.rst

testinfra/modules/iptables.py
```diff
@@ -0,0 +1,50 @@
...
+        rules = self.get_rules()
+        return self.rule in rules
+
+    def get_rules(self):
+        rules = []
```
philpep: I'd be nice to have `get_rule(self, chain=None)` returning rules for a specific chain, for instance `host.iptable_rule().get_rules('FORWARD')` would return the output of `iptables -S FORWARD`

Also consider adding docstring with example of `get_rule` usages.

scootersmk:
Rewrote things a bit to handle specifying a chain. Here is my thinking on the two ways this can be used(demonstrated by the two different test blocks:
  1. Check for existence of a specific rule, optionally specifying a table. Since the rule contains the chain at the beginning, I thought passing 'chain' to get_rule here seemed redundant and made the arguments overly complex. 
  2. Pull a list of all rules that optionally match a specific table and/or chain. Then the user can perform as many tests on that list as they wish.

Updated docs/formatting to make all tests happy. I'm ready for this to be merged if you are.

testinfra/modules/iptables.py
scootersmk: fixed

testinfra/modules/iptables.py
scootersmk: fixed, let me know if you had something else in mind. See explanation in general discussion.

philpep:
Sorry to not have spoted in my first review, but I think this module could have a better API.

What do you think of this api instead ?

```python
rules = host.iptables.rules(table='filter', chain='INPUT')
assert '-A INPUT -p tcp --dport http -j ACCEPT' in rules
assert '-A INPUT -j LOG' in rules
```

Where `rules(table=None, chain=None)` would simply return iptables rules as a python list.
You can inherit from `testinfra.modules.InstanceModule` to have `host.iptables` resolve to an instance of `Iptables` directly.

I find it's more simple (a single method `rules()` is needed), more pythonic (use of `in` operator) and less potential performance issues than with the `exists()` API (since each call will call a new `get_rules()`).

Tell me if you are ok with this API and if you have time to implement it (I you have no time, I can implement it based on your work).

Thanks, that's actually a big contribution!

testinfra/modules/iptables.py
```diff
@@ -0,0 +1,87 @@
...
+        ).exists
+        True
+
+        """
+
+        rules = self.get_rules()
+        return self.rule in rules
+
+    def get_rules(self, table='filter', chain=None):
+        """Returns list of iptables rules by running
```
philpep: Looks like the `by running` is the beginning of an unfinished sentence, I'm wrong ?

scootersmk: fixed

scootersmk:
Yes I agree with that API design as its both simple and flexible. Take a look at the most recent commit.

philpep:
Awesome, merged, thanks!

PR #273, opened 2017-11-23
Fix typos and style for backends.rst

marco-m:
Among the changes, I removed the references to ` --sudo` for the various backends because that option is already documented at the beginning of the page.


philpep:
Merged, thanks @marco-m !

Issue #247, opened 2017-09-12
remote test does not create TTY - can't use sudo


brucellino:
Hi there.
I'm using testinfra to test servers over ssh. The connection user has sudo, but it seems that the shell that is assigned to it does not allocate a tty, so the test fails.

```
_________________________________________________________________ test_track_and_trace[ssh://ci.sagrid.ac.za] __________________________________________________________________

host = <testinfra.host.Host object at 0x7f4edb550810>

    def test_track_and_trace(host):
>       config_file = host.file('/etc/httpd/conf.d/jenkins.conf')

ci.py:27: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/local/lib/python2.7/dist-packages/testinfra/host.py:91: in __getattr__
    obj = module_class.get_module(self)
/usr/local/lib/python2.7/dist-packages/testinfra/modules/base.py:22: in get_module
    klass = cls.get_module_class(_host)
/usr/local/lib/python2.7/dist-packages/testinfra/modules/file.py:176: in get_module_class
    if host.system_info.type == "linux":
/usr/local/lib/python2.7/dist-packages/testinfra/modules/systeminfo.py:122: in type
    return self.sysinfo["type"]
/usr/local/lib/python2.7/dist-packages/testinfra/modules/systeminfo.py:31: in sysinfo
    self._sysinfo = self.get_system_info()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <testinfra.modules.base.SystemInfo object at 0x7f4edb50b1d0>

    def get_system_info(self):
        sysinfo = {
            "type": None,
            "distribution": None,
            "codename": None,
            "release": None,
        }
[...]

PR #275, opened 2017-11-29
Use the RequireTTY setting if specified in a provided SSHConfig

jbmchuck:
Currently the ssh_config option RequireTTY is not used when using the
paramiko backend with an ssh config file.  This option is useful
when using sudo and connecting to hosts which have the requiretty
option set in sshd_config. Setting requiretty in sshd_config is no
longer a common default, but it is frequently on older distributions and
in some 'hardened' distributions.

This change causes paramiko to use Channel.get_pty() to request a
pseudo-terminal if the RequireTTY option is specified in a provided
ssh_config.

This allows addressing

    https://github.com/philpep/testinfra/issues/247

without modifying the server-side configuration.


philpep:
Merged, thanks!

PR #276, opened 2017-12-01
add alpine package

akabella1:


testinfra/modules/package.py
```diff
@@ -146,3 +149,20 @@ def version(self):
     def release(self):
         return self.check_output('rpm -q --queryformat="%%{RELEASE}" %s',
                                  self.name)
+
+
+class AlpinePackage(Package):
+
+    @property
+    def is_installed(self):
+        return self.check_output("apk -e info %s", self.name) == self.name
```
philpep: I think this doesn't works for a non installed package since `apk` will exit with a status code 1 and will cause `check_output()` to raise an assertion.

I think this should be `return self.run_test('apk info -e %s', self.name).rc == 0`

testinfra/modules/package.py
```diff
@@ -146,3 +149,20 @@ def version(self):
...
+
+    @property
+    def version(self):
+        out = self.check_output("apk -e -v info %s", self.name).split("-")
```
philpep: This will not work for packages containing a `-` (like `libc-utils`).
This could be fixed with `out = self.check_output(...); out = out[len(name)+1:].split('-', 1)[0]`

testinfra/modules/package.py
```diff
@@ -146,3 +149,20 @@ def version(self):
...
+        return out[-2]
+
+    @property
+    def release(self):
+        out = self.check_output("apk -e -v info %s", self.name).split("-")
```
philpep: `out = out[len(name)+1:].split('-', 1)[1]`

akabella1:
Hellow, is_installed function really was not done correctly

But version and release do not agree, they will work with packages containing "-" because because the list is addressed from the end, 
the output of the command contains the structure
name-version-release
and the first from the end out[-1] it is release
second from the end out[-2] it is version, regardless of package name
```
@property
def version(self):
out = self.check_output("apk -e -v info %s", self.name).split("-")
return out[-2]
```
```
@property
def release(self):
out = self.check_output("apk -e -v info %s", self.name).split("-")
return out[-1]
```

>Воскресенье,  3 декабря 2017, 8:14 +07:00 от Philippe Pepiot <notifications@github.com>:
>
>@philpep requested changes on this pull request.
>Hi, thanks for your contribution !
>I've added some comments that need to be fixed.
>Also please remove the unwanted file with git rm .idea/vcs.xml
>I would be awesome if you could add some tests (add a new image in images/)
>The test could be:
>@pytest.mark.testinfra_hosts('docker://alpine')
>def test_alpine_package(host):
>    busybox = host.package('busybox')
>    assert busybox.is_installed
>    assert not host.package('nginx').is_installed
>    assert busybox.version == '1.27.2'
>    assert busybox.release == 'r6'
>----------------------------------------------------------------------
>In  testinfra/modules/package.py :
>> @@ -146,3 +149,20 @@ def version(self):
>     def release(self):
>         return self.check_output('rpm -q --queryformat="%%{RELEASE}" %s',
>                                  self.name)
>+
>+
>+class AlpinePackage(Package):
>+
>+    @property
>+    def is_installed(self):
>+        return self.check_output("apk -e info %s", self.name) == self.name
>
>I think this doesn't works for a non installed package since apk will exit with a status code 1 and will cause check_output() to raise an assertion.
>I think this should be return self.run_test('apk info -e %s', self.name).rc == 0
>----------------------------------------------------------------------
>In  testinfra/modules/package.py :
>> @@ -146,3 +149,20 @@ def version(self):
>     def release(self):
>         return self.check_output('rpm -q --queryformat="%%{RELEASE}" %s',
>                                  self.name)
>+
>+
>+class AlpinePackage(Package):
>+
>+    @property
>+    def is_installed(self):
>+        return self.check_output("apk -e info %s", self.name) == self.name
>+
>+    @property
>+    def version(self):
>+        out = self.check_output("apk -e -v info %s", self.name).split("-")
>
>This will not work for packages containing a - (like libc-utils).
>This could be fixed with out = self.check_output(...); out = out[len(name)+1:].split('-', 1)[0]
>----------------------------------------------------------------------
>In  testinfra/modules/package.py :
>> +
>+
>+class AlpinePackage(Package):
>+
>+    @property
>+    def is_installed(self):
>+        return self.check_output("apk -e info %s", self.name) == self.name
>+
>+    @property
>+    def version(self):
>+        out = self.check_output("apk -e -v info %s", self.name).split("-")
>+        return out[-2]
>+
>+    @property
>+    def release(self):
>+        out = self.check_output("apk -e -v info %s", self.name).split("-")
>
>out = out[len(name)+1:].split('-', 1)[1]
>—
>You are receiving this because you authored the thread.
>Reply to this email directly,  view it on GitHub , or  mute the thread .


С уважением,
Кирилл Чуприн
95.kirillchuprin@mail.ru

testinfra/modules/package.py
akabella1: version do not agree, they will work with packages containing "-" because because the list is addressed from the end, 

the output of the command contains the structure
name-version-release
and the first from the end out[-2] it is version

testinfra/modules/package.py
akabella1: release do not agree, they will work with packages containing "-" because because the list is addressed from the end, 

the output of the command contains the structure
name-version-release
and the first from the end out[-1] it is release

testinfra/modules/package.py
akabella1: is_installed function really was not done correctly

testinfra/modules/package.py
philpep: Indeed, missed that! It's ok then.

PR #278, opened 2017-12-05
228 support alpine package system

akabella1:


philpep:
Oops, the PR was closed unexpectedly while modifying it, re-opened in #279

I dropped some unwanted code (like accepting 127 as a valid return code for testing if command exists...). I think we don't need to test whole sshd / service and so on, this can be done in a separate PR eventually.

PR #280, opened 2017-12-08
228 add support for alpine package manager

akabella1:
Add support package and test for them


philpep:
Ok thanks and sorry for being late on this. I've added a patch to drop unnecessary setup of the alpine image that is not needed for tests. Will merge as soon as the CI is green.

philpep:
Merged, thanks!

PR #282, opened 2018-01-17
Change check_all to check_call in doc example

cptavatar:
I was experimenting with the docker examples and noticed that there appeared to be typo - it looks like subprocess.check_all(...) should actually be subprocess.check_call(...)


philpep:
Merged, thanks for catching and fixing this!

PR #283, opened 2018-01-20
Add service, process, and systeminfo support for Alpine

### LINKED ISSUE #228: Support alpine docker images
My Dockerfile starts with:

`FROM nginx:1.12-alpine`

In my Dockerfile I want to pull images based on Alpine because these are way smaller than Ubuntu, Centos, and the like. Alpine uses BusyBox so command options are limited. 

[BusyBox Commands]https://busybox.net/downloads/BusyBox.html#commands )

jharley:
This should resolve #228

### DIFF
--- a/testinfra/modules/process.py
+++ b/testinfra/modules/process.py
@@ -110,7 +110,9 @@ class Process(InstanceModule):
 
     @classmethod
     def get_module_class(cls, host):
-        if (host.system_info.type == "linux"
+        if host.file("/bin/ps").linked_to == "/bin/busybox":
+            return BusyboxProcess
+        elif (host.system_info.type == "linux"
                 or host.system_info.type.endswith("bsd")):
             return PosixProcess
         else:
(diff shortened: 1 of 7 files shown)


jharley:
@philpep the bulk of the pylint errors are in code I didn't touch, did you want me to fix them (or ignore them in pylintrc) to get this to "go green"?  There's one from my changes (7 returns in a method instead of the previous 6), which I can resolve also

jharley:
Actually, I just saw that the `is_running` for the OpenRC service returned an 8(?!), which doesn't happen locally.  Will try to reproduce.

testinfra/modules/process.py
```diff
@@ -110,7 +110,9 @@ def _get_process_attribute_by_pid(self, pid, name):
 
     @classmethod
     def get_module_class(cls, host):
-        if (host.system_info.type == "linux"
+        if host.system_info.distribution == "alpine":
+            return BusyboxProcess
```
philpep: Is it also possible to have procps installed in alpine ?
Maybe we should detect busybox with `if host.file('/bin/ps').linked_to == '/bin/busybox')` ?

philpep:
Hi, thanks for contributing ! Your changes looks almost good to me (see minor remark on busybox handling).
pylint issues seems related to a recent pylint upgrade, I'll fix them asap, you can just ignore them for now.

testinfra/modules/process.py
jharley: That is a much nicer method of detection of Busybox!  I will make that change.

It is possible to have "procps" installed on Alpine, but it isn't there by default I don't believe. 

The test image does currently have it installed (left around in my hacking about), which means  `BusyBoxProcess` is not being properly called or tested.  I will remove "procps" from the Alpine image and make any changes required.

jharley:
@philpep Please let me know if any more changes are required or if I can do anything more to help?

philpep:
@jharley sorry for the delay. Your changes looks good to me, I'll merge as soon as the CI is green. thanks!

Issue #264, opened 2017-11-08
Feature Request: Docker module


blackillzone:
Is this possible to add a module, to check if a docker service is up (like the service module would currently do) ?

I would see something like this (with more or less functionalities):

```
class Docker(name)
     is_running
Test if container is running

     is_listening
Test if container is listening to ports

     version
Return version of the image that the container is running
```

What do you think about this ? Could it be relevant to add it ?

sashgorokhov:
Sounds interesting, I could submit a PR for that, though I am not currently familiar with testinfra, so it could take some time. @philpep, what do you think?

philpep:
I think this could be useful to have a docker module. Probably ansible or salt source code could be an inspiration.

mludy:
Hi Guys, what is the status of it ? Having docker module would be very useful with dockerized environments test

ecks:
please take a look at pull request #300

Kilo59:
@blackillzone I think this issue can be closed. The PR mentioned above has been already been accepted and merged.

PR #300, opened 2018-04-04
Docker module

ecks:
this pull request is an attempt at implementing docker module, which allows you to test whether or not certain docker containers are running on a particular connection backend. 

this pull request references issue #264


ecks:
looks like the failing travis unit test is unrelated. i tested it in my local env, sometimes it passes, other times it does not.

testinfra/modules/docker.py
```diff
@@ -0,0 +1,76 @@
...
+ 
+    Example:
+
+    >>> host = testinfra.get_host('local://')
+
+    >>> nginx = host.docker("nginx")
+    >>> assert nginx.is_running
+
+    >>> nginx.is_listening = ['8080', '8081']
+    >>> assert nginx.is_listening
```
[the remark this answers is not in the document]
ecks: yeah that is cleaner, thanks! changed the code based on your comment.

ryanjfrizzell:
bump

sagikazarmark:
It would be nice to see this feature merged!

sagikazarmark:
Is there a way to add this to my setup while we are waiting for merging/releasing this?

ecks:
Hello,

Thank you for showing interest in the pull request. If you want to use the code from this pull request right away, you can do a couple of different things. Easiest thing would be to

git clone https://github.com/ecks/testinfra
git checkout docker_module

That will allow you to get access to the code, you'll be able to run it, but will not be able to push any changes (not unless I change permissions of that repo). 

Otherwise, you can hit "fork" on my repo, and it will create a matching repo under your username, which you will have full access to. Then you would clone your repo and switch branches the same way.




> Is there a way to add this to my setup while we are waiting for merging/releasing this?

ecks:
@philpep Can you look over my pull request once again?

testinfra/modules/docker.py
```diff
@@ -0,0 +1,72 @@
...
+        run_ports = self.running_ports()
+        assert all(i in ports for i in run_ports)
+        return True
+
+    def versions(self):
+        image = self.running_image()
+        cmd = self.run("docker images %s --format '{{.Tag}}'", image)
+        return cmd.stdout.splitlines()
+
+    @property
```
philpep: Hi @ecks , I'm really sorry to have been sooo long to review this...

I'm not convinced by `version()`, `is_listening()`, `running_ports()` and `running_image()`. I think I can merge the docker module with just `Docker(name).is_running` which seems definitively useful. Can you update your PR for this, so we can move forward ?

For other features, I think we can add a helper "inspect()" which return the output of `docker inspect` as python dict object. So users can easily make assertion on it, what do you think ?

Thanks!

testinfra/modules/docker.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+
+    def __init__(self, name):
+        self.name = name
+        super(Docker, self).__init__()
+
+    @property
+    def is_running(self):
+        cmd = self.run("docker ps -q -f ancestor=%s --format '{{.Image}}'",
+                       self.name)
```
philpep: Sorry missed that on first review. I think testing "image" is running is wrong because you can have multiple running containers with the same image, also if a new ``nginx:latest`` is pulled I'm not sure ``{{.Image}}`` will remain the same.

I'd rather parse the output of ``docker ps --format '{{ .ID }} {{ .Names }}'`` and find ``self.name`` in the list of docker ids and names.

ecks: Hey @philpep I wasn't sure what to exactly check for. It could be debatable that sometimes you might want to know if an instance of an image is running and want to assert on that, especially if you know the name of the image doesn't change. If you are asserting on the name of the running container, i.e. sleepy_hare, you would have to change your code each time you spawn a different one.

It would make sense though to assert on the running container name though, if that container is going to exist for a while. I changed the code to reflect that, based on your suggestions. Now you should be able to perform the actions you specified, additionally you can get the image id by using

```nginx.image_id```

I think for future we can implement another class called "docker_image" that will represent the image info, since "docker inspect image_id" returns some useful stuff as well. With those two classes you should be able to do so something like

```assert len(host.docker_image("nginx:latest").running_instances) > 0```

Comment (philpep) [reply testinfra/modules/docker.py]: What about adding a classmethod to return the list of `Docker` instances running ? Eventually with a parameter to filter containers.

```
>>> host.docker.get_containers()
[Docker('XXX'), Docker('YYY')]
>>> host.docker;get_containers(image="nginx:latest")
[Docker("XXX")]
```

I think it's better to do this in a following PR, thanks for updating this PR, I think I can merge part of it, that will give us a good base to work on.

File: testinfra/modules/docker.py [hunk 4]
```diff
@@ -0,0 +1,75 @@
...
+    def inspect(self):
+        cmd = self.run("docker inspect %s", self.inst_name)
+        json_out = json.loads(cmd.stdout)
+
+        if len(json_out) == 1:
+            return json_out[0]
+
+    @property
+    def id(self):
+        return self.inspect()["Id"][:12]  # display only first twelve chars
```
philpep: Why twelve chars ?

philpep: I think it's better to let the called to this if needed.

philpep:
Hi, I finally merged part of this PR with some changes:

* Reuse inspect() for is_running
* inst_name --> _name
* drop image_id (I'm not sure it should work like this for now)
* return all chars in "id" property
* use check_output() in inspect() and do not retun None when docker inspect return [] (do not hide errors)
* add docker module to the documentation

Unfortunatelly I was not able to push on your repo, so I amended all changes in a single commit (and still keeping your authorship) https://github.com/philpep/testinfra/commit/b741de52d34892d5f1e2331749b798948268e379

Thanks for your work on this !

As a followup, I'll be happy to have a new PR implementing the `get_containers()` I mentioned before.

ecks:
Ok thanks! Yeah I could take a look at *get_containers()* in the future.

Kilo59:
@philpep Do your docs build automatically? Because I don't see the latest version in the changelog, or the modules section.
https://testinfra.readthedocs.io/en/latest/changelog.html
https://testinfra.readthedocs.io/en/latest/modules.html

philpep:
@Kilo59 thanks for reporting this ! It's supposed to build automatically, I'll check this ASAP.

Kilo59:
@ecks is it possible to use this with a remote Docker Host?

PR #301, opened 2018-04-05
Make PipPackage compatible with pip 10.x

acrewdson:
Recently-released 10.x versions of pip use a different output format by default for `pip list` (columns). Here we are switching to specify that format explicitly, making the code compatible with both 9.x and 10.x versions of pip.

### DIFF
--- a/testinfra/modules/pip.py
+++ b/testinfra/modules/pip.py
@@ -21,6 +21,27 @@ from testinfra.modules.base import InstanceModule
 class PipPackage(InstanceModule):
     """Test pip packages status and version"""
 
+    @staticmethod
+    def _first_match(line, regexes):
+        for r in regexes:
+            match = r.match(line)
+            if match is not None:
+                return match
+        raise RuntimeError("could not parse {0}".format(repr(line)))
+
+    def _pip_list(self, pip_path, outdated=False):
+        extras = " -o " if outdated else " --no-index "
+        out = self.run_expect(
+            [0, 2],
+            "{} list {} --format=columns".format(pip_path, extras)
+        )
+        if out.rc == 0:
+            lines = out.stdout.splitlines()[2:]  # drops column headers
+        else:  # older version of pip, we assume
+            out = self.check_output("{} list {}".format(pip_path, extras))
+            lines = out.splitlines()
+        return lines
+
     def get_packages(self, pip_path="pip"):
         """Get all installed packages and versions returned by `pip list`:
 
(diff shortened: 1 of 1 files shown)


testinfra/modules/pip.py
```diff
@@ -29,14 +29,16 @@ def get_packages(self, pip_path="pip"):
          'mywebsite': {'version': '1.0a3', 'path': '/srv/website'},
          'psycopg2': {'version': '2.6.2'}}
         """
-        output_re = re.compile(r'^(.+) \((.+)\)$')
+        output_re = re.compile(r'^(.+?)\s+(.+)$')
         pkgs = {}
-        out = self.check_output("%s list --no-index", pip_path)
-        for line in out.splitlines():
+        out = self.check_output(
+            "%s list --no-index --format=columns", pip_path
```
philpep: Thanks for the patches, they work well.

But I just noticed that starting with pip 9 there is a `--format=json` option which will make things easier for us.
Also it's not a strong opinion but I'd like to keep compatibility with earlier pip version (those found on debian wheezy or centos 6). I think we can write something like this:

```
out = self.run_expect([0, 2], "pip list --no-index --format=columns")
if out.rc == 0:
    data = json.loads(out.stdout)
    # continue for pip >= 9
else:
    out = self.check_output("pip list --no-index")
    # continue for pip < 9
```

What do you think ?

acrewdson: @philpep Thanks for the review.

It's funny you mention the `json` option, since I almost used that, then thought I'd keep it simple and stick with the text-based output. It could be a nice improvement for the future, but this seemed like a quicker route to regain pip 10.x compatibility.

Regarding the earlier pip versions, that seems good to me, but how do we test that? All the tests seemed to pass for my change, so does that mean we need to add some extra tests for the old distros you mentioned above?

philpep: I think it's ok to not write tests for old pip version, you can just test manually for pip 8 and pip 7 and It will be fine to me.

acrewdson: 👍 I'll work on that and get back to you, then. Thanks.

acrewdson:
@philpep I updated the patch based on your feedback. Please take another look when you have a chance, thanks!

philpep:
Merged with some changes to use `--format=json` for pip >= 9.

Thanks!

acrewdson:
@philpep thank you!

PR #309, opened 2018-05-04
Use HTTPS protocol when specified in WinRM backend

cowdude:
Hello,

It seems there is a typo in the WinRM backend that prevents it to connect to WinRM services over HTTPS.

This change fixed the issue against our environment.

Remote host: AWS Windows server 2016 base 10, version 2018.03.24
Testinfra host: Ubuntu Linux 16.04, python 3.6

Example PowerShell script to configure WinRM service with HTTPS:  https://github.com/PeteGoo/packer-win-aws/blob/master/ec2-userdata.ps1


philpep:
Merged, thanks!

PR #317, opened 2018-05-22
Add namespace support for kubectl connector.

LarsT42:
usage: py.test --connection=kubectl --hosts=<pod-name>[/<image-name>][%<namespace>] ...

where namespace is the k8s namespace where you work inside.
So it's not necassary to switch the default namespace from your kubectl client installation
to access/work with a specific namespace

### DIFF
--- a/testinfra/backend/__init__.py
+++ b/testinfra/backend/__init__.py
@@ -46,9 +46,6 @@ def parse_hostspec(hostspec):
         url = urllib.parse.urlparse(hostspec)
         kw["connection"] = url.scheme
         host = url.netloc
-        path = url.path
-        if path:
-            kw['container'] = path[1:]
         query = urllib.parse.parse_qs(url.query)
         for key in ('sudo', 'ssl', 'verify_ssl'):
             if query.get(key, ['false'])[0].lower() == 'true':
--- a/testinfra/backend/base.py
+++ b/testinfra/backend/base.py
@@ -204,14 +204,6 @@ class BaseBackend(object):
             user, name = name.split("@", 1)
         return name, user
 
-    @staticmethod
-    def parse_podspec(podspec):
-        name = podspec
-        container = None
-        if "/" in name:
-            name, container = name.split("/", 1)
-        return name, container
-
     def get_encoding(self):
         cmd = self.run(
             "python -c 'import locale;print(locale.getpreferredencoding())'")
--- a/testinfra/backend/kubectl.py
+++ b/testinfra/backend/kubectl.py
@@ -21,41 +21,24 @@ class KubectlBackend(base.BaseBackend):
     NAME = "kubectl"
 
     def __init__(self, name, *args, **kwargs):
-        self.container = None
-        self.name, self.container = self.parse_podspec(name)
-
-        if kwargs:
-            if "namespace" in kwargs:
-                self.namespace = kwargs.get('namespace')
-            if "container" in kwargs:
-                self.container = kwargs.get('container')
-
-        if "/" in self.name:
-            self.name, self.container = self.name.split("/", 1)
-
+        self.name = name
+        self.container = kwargs.get('container')
+        self.namespace = kwargs.get('namespace')
         super(KubectlBackend, self).__init__(self.name, *args, **kwargs)
 
     def run(self, command, *args, **kwargs):
         cmd = self.get_command(command, *args)
         # `kubectl exec` does not support specifying the user to run as.
         # See https://github.com/kubernetes/kubernetes/issues/30656
-        if self.container is None:
-            if self.namespace:
-                out = self.run_local(
-                    "kubectl -n %s exec %s -- /bin/sh -c %s",
-                    self.namespace, self.name, cmd)
-            else:
-                out = self.run_local(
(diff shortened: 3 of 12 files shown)


LarsT42:
Hi Philippe,
I changed the kubectl connector according to your proposals.

kindly regards
Lars

LarsT42:
Hi Philippe,
I changed the kubectl connector according to your proposals

kindly regards
Lars

testinfra/backend/__init__.py
```diff
@@ -46,11 +46,14 @@ def parse_hostspec(hostspec):
         url = urllib.parse.urlparse(hostspec)
         kw["connection"] = url.scheme
         host = url.netloc
+        path = url.path
+        if path:
+            kw['path'] = path
```
philpep: This `path` looks unused.

testinfra/backend/base.py
```diff
@@ -204,6 +204,14 @@ def parse_containerspec(containerspec):
             user, name = name.split("@", 1)
         return name, user
 
+    @staticmethod
+    def parse_podspec(podspec):
+        name = podspec
+        container = None
+        if "/" in name:
+            name, container = name.split("/", 1)
+        return name, container
```
philpep: This could be simplier:

```
if '/' in podspec:
    return podspec.split('/', 1)
return podspec, None
```

testinfra/backend/kubectl.py
```diff
@@ -21,23 +21,41 @@ class KubectlBackend(base.BaseBackend):
...
     def __init__(self, name, *args, **kwargs):
-        self.name, self.user = self.parse_containerspec(name)
+        self.container = None
+        self.name, self.container = self.parse_podspec(name)
+
+        if kwargs:
+            if "namespace" in kwargs:
+                self.namespace = kwargs.get('namespace')
+            if "container" in kwargs:
+                self.container = kwargs.get('container')
```
philpep: ``parse_containerspec()`` also handled the user defined with ``user@pod``, but I see that ``kubectl`` doesn't allow to exec as arbitrary user, so I'm ok with your change.

If a user provide container in both way: ``pod/a?container=b`` we will use one of them but it's unclear to me, I'd prefer to raise or not parse one of the container (which will lead to raise in kubectl).

So I think ```__init__``` could be:

```python
self.namespace = kwargs.get('namespace')
self.container = kwargs.get('container')
if self.container is None:
    self.name, self.container = self.parse_podspec(name)
else:
   self.name = name
super(KubectlBackend, self).__init__(self.name, *args, **kwargs)
````

File: testinfra/backend/kubectl.py [hunk 4]
```diff
@@ -21,23 +21,41 @@ class KubectlBackend(base.BaseBackend):
...
+
         if "/" in self.name:
             self.name, self.container = self.name.split("/", 1)
-        else:
```
philpep: This need part need to be dropped I think.

testinfra/backend/kubectl.py
```diff
@@ -21,23 +21,41 @@ class KubectlBackend(base.BaseBackend):
...
-                "kubectl exec %s -c %s -- /bin/sh -c %s",
-                self.name, self.container, cmd)
+            if self.namespace:
+                out = self.run_local(
+                    "kubectl -n %s exec %s -c %s -- /bin/sh -c %s",
+                    self.namespace, self.name, self.container, cmd)
+            else:
+                out = self.run_local(
+                    "kubectl exec %s -c %s -- /bin/sh -c %s",
+                    self.name, self.container, cmd)
```
philpep: Instead of handling the 4 cases we could build a `kubectlcmd` string and a `cmd_args` array.

````python
kubectlcmd = 'kubectl '
cmd_args = []
if self.namespace is not None:
    kubectlcmd += '-n %s '
    cmd_args.append(self.namespace)
if self.container is not None:
    kubectlcmd += '-c %s '
    cmd_args.append(self.container)
kubectlcmd += 'exec %s -- /bin/sh -c %s'
cmd_args.extend([self.name, cmd])
out = self.run_local(kubectlcmd, *cmd_args)
```

testinfra/backend/__init__.py
```diff
@@ -46,11 +46,14 @@ def parse_hostspec(hostspec):
         url = urllib.parse.urlparse(hostspec)
         kw["connection"] = url.scheme
         host = url.netloc
+        path = url.path
+        if path:
+            kw['container'] = path[1:]
```
philpep: Oh, I see that ``urlparse()`` splitted the part after `/` into this `path` attribute.
So this has never worked.

```
>>> import testinfra
>>> backend = testinfra.get_host('kubectl://pod/c').backend
>>> backend.name
'pod'
>>> backend.container is None
True
```

In this case I think you can just drop this feature of splitting the container and the pod name with `/` and just rely on url querystring.

A good test could be:

````
b = testinfra.get_host('kubectl://pod?namespace=space&container=nginx').backend
assert b.name == 'pod'
assert b.namespace == 'space'
assert b.container == 'nginx'
```

Also I think this deserve some documentation in `doc/source/backend.rst` ;)

philpep:
Merged with minor fixes. Thanks!

PR #318, opened 2018-05-23
allow calling salt with sudo

gtmanfred:


philpep:
Ok thanks, there's also a ``--sudo`` option when calling `py.test`, but both (sudo option / sudo modules) might have bugs with salt module or salt backend, so let me know if you think you catches one of these :)

PR #319, opened 2018-05-23
Share runners (and parsed inventory) among hosts

sethp-nr:
Presently, the caching applied to `AnsibleRunners` is only effective at
the host level -- meaning that while we'll only check the inventory once
per host, we cannot do it less often. With dynamic inventories that can
take multiple seconds to compute their value, this cost can become
expensive quickly.

This change moves the cache to the class level, so that we'll execute
the ansible inventory once per `ansible_inventory` argument and no more
often.

NB: if a client calls into AnsibleBackend using a relative path, changes
the working directory, and calls `get_hosts` again with the same
relative path, they'll get stale results.

That workflow seems unlikely, but it might be nice to at least log a message on cache hit so we can shorten the time to resolution in that situation – I haven't been successful in emitting a log message in AnsibleBackend that a person using `py.test` will see. Any tips?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,12 @@
 Changelog
 =========
 
+1.13.1
+======
+
+* package: fix is_installed and version behavior for uninstalled packages (#321 and #326)
+* ansible: Use predictibles test ordering when using pytest-xdist to fix random test collections errors (#316)
+
 1.13.0
 ======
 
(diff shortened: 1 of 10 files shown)


philpep:
I modified your commits to move the cache to AnsibleRunner class.

Merged, thanks!

sethp-nr:
@philpep ah, my apologies – I've been away for the last week. Thank you much for fixing up the PR and merging!

PR #320, opened 2018-05-25
[WIP] Fix unix socket testing with ss

azaghal:
This is the initial, work-in-progress fix for socket testing via `ss` on GNU/Linux systems.

The main catch is that the tests I've implemented for the change won't work on Docker, since the Python code gets called on Docker host, and not on containers.

But... I would be interested in making the tests work on containers directly, as a way to keep regressions at bay, as well as to make the tests work in such a way that both `netstat` and `ss` code paths get tested.

Figured I might actually try to create a PR just to get some feedback instead of sitting on code until it's all sorted out.

I hope you won't mind too much (but feel free to point out you do, no hard feelings, I'm still interested in landing the fix :)


azaghal:
Sorry, managed to press enter before double-checking writing etc :)

philpep:
Hi, thanks for contributing!

I'm not sure about what you're trying to fix  in the ss module ?
You're right, as I read your new test file, and given how testinfra self-tests are run, the `host` refer to a running `images/debian_stretch` docker image, so It cannot work (you bind socket in the docker host running tests, and test using `ss` inside the docker container).

azaghal:
The issue I faced was when using the `Socket.is_listening` test with a unix socket. Even if socket file exists at specified path, and an application is listening on it, the `is_listening` will always return `False`. Running the identical test using the `netstat` backend would work correctly, and return `True`.

If you happen to have a local machine with `ss`, I think you should be able to reproduce the error using the provided tests.

Apropos tests, I could remove them, but if you have a good suggestion on how to make them actually run within Docker images (I guess mainly mechanism on how to spawn some test servers to listen on TCP/UDP/Unix socket prior to running tests), I'm open to work towards it :)

azaghal:
@philpep Anything else I could do to get this PR going? :) Did my explanation on broken use of `ss` make any sense?

philpep:
Hi @azaghal , sorry for my late reply. Ok I understand the bug with unix sockets now and the way you fix it looks good to me. I'll be quite complex for now to integrate your added tests, I think #333 might help here (where we could bind sockets using netcat utility in a background command).

I pushed some changes in the PR (drop the test_socket.py file and add a is_listening check on a unix socket that already bind inside the docker image) and I'll merge after travis CI results. Thanks!

philpep:
Merged, thanks!

Issue #303, opened 2018-04-09
Implement __eq__ for File objects


mdelapenya:
Hello, I'm getting this error while checking the `linked_to` of a valid symlink.

```python
        def test_paths_are_symliks(host):
            myPath = host.file("/mnt/jenkins/workspace")
            myLink = host.file("/home/ubuntu/workspace")
            assert myPath.exists
            assert myLink.exists
            assert myLink.is_symlink
            assert myLink.linked_to == myPath

    E           AssertionError: assert '/mnt/jenkins/workspace' == <file /mnt/jenkins/workspace>
    E            +  where '/mnt/jenkins/workspace' = <file /home/ubuntu/workspace>.linked_to
```

Is the last assertion returning a file object instead of the string representation of its path?

Thanks!

PS: close this issue if you consider is not a bug

mdelapenya:
I totally see it now...  I don't why I did not see it before: I was just putting the eyes in on side of the equation, when the problem was at the other 😄

It works now, thanks!

PR #328, opened 2018-05-29
fix issue #303

barnabasJ:
Implements \_\_eq\_\_ and \_\_ne\_\_ methods on the File object.
If both objects are files, the paths are getting compared,
if the other object is a string, it's compared with the path directly.

### DIFF
--- a/testinfra/modules/file.py
+++ b/testinfra/modules/file.py
@@ -171,6 +171,16 @@ class File(Module):
     def __repr__(self):
         return "<file %s>" % (self.path,)
 
+    def __eq__(self, other):
+        if isinstance(other, File):
+            return self.path == other.path
+        elif isinstance(other, str):
+            return self.path == other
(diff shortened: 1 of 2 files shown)


testinfra/modules/file.py
```diff
@@ -171,6 +171,16 @@ def size(self):
     def __repr__(self):
         return "<file %s>" % (self.path,)
 
+    def __eq__(self, other):
+        if isinstance(other, File):
+            return self.path == other.path
+        elif isinstance(other, str):
+            return self.path == other
```
philpep: I'd use `isinstance(other, six.string_types)` that will also match python 2 unicode string (``isinstance(u'foo', str)`` return ``False`` with python2).

Otherwise your change looks good to me, thanks for implementing this!

philpep:
Merged, thanks!

Issue #327, opened 2018-05-28
SysvService has a subtle flaw if rc?.d directories are symlinks


codylane:
While testing this afternoon, I noticed another small issue if a system under test uses the `SysvService` to check for service state.  This one is very subtle and didn't catch my eye right away.

https://github.com/philpep/testinfra/blob/master/testinfra/modules/service.py#L93-L98

could be changed to this which follows symlinks incase the `rc?.d` are symlinks because on some Linux machines they are indeed symlinks.  

```
diff --git a/testinfra/modules/service.py b/testinfra/modules/service.py
index 7b3b197..ad1b7e0 100644
--- a/testinfra/modules/service.py
+++ b/testinfra/modules/service.py
@@ -133,7 +133,7 @@ class SysvService(Service):
     @property
     def is_enabled(self):
         return bool(self.check_output(
-            "find /etc/rc?.d/ -name %s",
+            "find -L /etc/rc?.d/ -name %s",
             "S??" + self.name,
         ))

```

Here is an example of a Linux box with `rc?.d` symlinks.  The change above should behave the same with or without symlinks.  Without the patch, `is_enabled` would always return `False`.

`cat /etc/redhat-release`
```
CentOS release 6.9 (Final)
```

` ls -ld /etc/rc?.d`
```
lrwxrwxrwx 1 root root 10 Apr  2 18:48 /etc/rc0.d -> rc.d/rc0.d
lrwxrwxrwx 1 root root 10 Apr  2 18:48 /etc/rc1.d -> rc.d/rc1.d
lrwxrwxrwx 1 root root 10 Apr  2 18:48 /etc/rc2.d -> rc.d/rc2.d
lrwxrwxrwx 1 root root 10 Apr  2 18:48 /etc/rc3.d -> rc.d/rc3.d
lrwxrwxrwx 1 root root 10 Apr  2 18:48 /etc/rc4.d -> rc.d/rc4.d
lrwxrwxrwx 1 root root 10 Apr  2 18:48 /etc/rc5.d -> rc.d/rc5.d
lrwxrwxrwx 1 root root 10 Apr  2 18:48 /etc/rc6.d -> rc.d/rc6.d
```

`ls -ld /etc/rc.d`
```
drwxr-xr-x 10 root root 4096 Apr  2 18:49 /etc/rc.d
```

`ls -l /etc/rc.d`
```
drwxr-xr-x 2 root root 4096 Apr  2 18:49 init.d
[...]

Issue #243, opened 2017-09-07
Testinfra is choosing Upstart for CentOS 6 as service provider


codylane:
Hello,

I've been tracking down some failing tests and I've noticed that testinfra `1.6.5` is picking up the `Upstart` service provider for my test case for CentOS 6, which feels wrong and it should be choosing the `SysvService` instead.  

Here's an example:

```
    @pytest.mark.docker_images('centos:6')
    def test_httpd_service_starts_up_through_etc_initd(host, wait_for_svc):
        svc = host.run('/etc/init.d/httpd start')
        wait_for_svc(host, 'start', timeout=5)
    
        assert svc.rc == 0
>       assert host.service('httpd').is_running
E       AssertionError: assert False
E        +  where False = <service httpd>.is_running
E        +    where <service httpd> = <class 'testinfra.modules.base.UpstartService'>('httpd')
E        +      where <class 'testinfra.modules.base.UpstartService'> = <testinfra.host.Host object at 0x7fd013d85310>.service
```

philpep:
This is probably because you have an `initctl` command available in your `$PATH`. Is this command related to upstart ? Maybe we should enhance upstart detection, for instance the presence of `/etc/init/` directory and or present of the `status` command.

codylane:
Hey @philpep - Thanks for the reply. :)  Opps, yes, I forgot to explain it is happening because 'initctl' for some reason is in my path and problem pulled in via the `git` package on my system under test.  I monkey patched service for now in my `tests/conftest.py` which seems to work but it probably needs some more thought you like you suggested.

Here's how I patched it:

```
# tests/patches.py

import testinfra.modules
from testinfra.modules.service import *

class PatchedService(Service):

    @classmethod
    def get_module_class(cls, host):
        if host.system_info.type == "linux":
            if (
                host.exists("systemctl")
                and "systemd" in host.file("/sbin/init").linked_to
            ):
                return SystemdService

            if host.exists("initctl") and host.system_info.distribution.lower() != 'centos':
                return UpstartService

            return SysvService

        if host.system_info.type == "freebsd":
            return FreeBSDService

        if host.system_info.type == "openbsd":
            return OpenBSDService

        if host.system_info.type == "netbsd":
            return NetBSDService

        raise NotImplementedError

testinfra.modules.get_module_class('service').get_module_class = PatchedService.get_module_classcat tests/patches.py
import testinfra.modules
from testinfra.modules.service import *

class PatchedService(Service):

    @classmethod
[...]

PR #329, opened 2018-05-30
Bugfix sysvservice

codylane:
The pull request addresses concerns in
- #327 
- #243

1. The refactor to `SysvService` `is_enabled` should now complain loudly if `/etc/rc?.d` is not a directory or symlink.
2. Add new `centos_6` test container to test new `SysvService` behavior
3. Perhaps slight increase in code coverage.
4. Address a subtle `tox.ini` issue when running the `ansible1` tox environment when system running tests is not a docker host with socket `unix:///var/run/docker.sock` and otherwise requires the user of the `DOCKER_*` environment variables.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,5 +18,3 @@ flake8.report
 junit*.xml
 doc/build
 .cache
-.pytest_cache/
-.vagrant/
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ flake8.report
 junit*.xml
 doc/build
 .cache
+.pytest_cache/
+.vagrant/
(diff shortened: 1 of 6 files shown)


codylane:
FYI - I introduced the EMPTY COMMIT to re-trigger the theory I had that the second build failed due to Travis-CI not because of something these commits introduced.

philpep:
Merged with some modifications:

- drop non-explained add to .gitignore
- drop some of your added tests that where already covered by test_sshd_running
- drop passenv DOCKER_HOST from this PR and commited it in 2a6788bef43e96b00e69b4d2c860bcef35f45a64

Thanks!

PR #342, opened 2018-07-09
Dns module

ruimoreira:
This pull requests tries to implement a simple check to determine if a DNS record exists, I am expanding this to perform more tests but for now basic test is just to check if the record exists.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ flake8.report
 junit*.xml
 doc/build
 .cache
+.idea
--- a/testinfra/modules/dns.py
+++ b/testinfra/modules/dns.py
@@ -1,0 +1,31 @@
+# coding: utf-8
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -18,3 +18,4 @@ flake8.report
 junit*.xml
 doc/build
 .cache
+.idea
```
philpep: What is this `.idea` file ? If it's related to your personal dev environment software, please use a global gitignore.

testinfra/modules/dns.py
```diff
@@ -0,0 +1,31 @@
...
+class Dns(Module):
+    """Test dns record exists """
+
+    def __init__(self, record=None):
+        self.record = record
+        super(Dns, self).__init__()
+
+    @property
+    def exists(self):
+        return self.run_expect([0, 1], "host %s", self.record).rc == 0
```
philpep: I think this module does not offer valuable feature for testinfra. But I also think a kind of dns module make sense in testinfra. Maybe we should start to see what feature offer similar software (serverspec, inspec and goss). I think we can discuss about the spec (= a list of example assertions) first and then implement it.

philpep:
Also, there is an attempt here: https://github.com/philpep/testinfra/pull/199

I propose to write some spec prior writing implementations on https://github.com/philpep/testinfra/issues/358

PR #350, opened 2018-07-24
Env + ApacheConfig Module

barnabasJ:
The ApacheConfig module relies on [apacheconfig](https://github.com/etingof/apacheconfig).
A PR for apacheconfig is open at the moment to add the last necessary changes to make this work.
I'll still want to provide some tests but the Envirnment I'm working in right know won't allow me to install needed requirements to execute the test. But I'll do that on the weekend when I'm working from home.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,12 @@ flake8.report
 junit*.xml
 doc/build
 .cache
+
+# venv
+bin
+inlude
+lib
+lib64
+pip-selfcheck.json
+.pytest_cache
+include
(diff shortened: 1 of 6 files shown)


philpep:
I, thanks for contributing and I'm very very sorry for taking so much time before giving you some feedback on this PR...
I'm not sure these new features are worth for testinfra, because it's just about testing file existence and parsing using an external library, I think this could already be done in a few lines in a test by using the `file` module. I'm gonna close this PR, fell free to arge about the need of new modules for this in an issue.

PR #353, opened 2018-08-07
Use HTTP with WinRM when user disables SSL

robertsigler:
I've noticed that testinfra always attempts to use HTTPS with the WinRM backend, even when I set the `no_ssl` query parameter. This is causing problems for me because the host in question is not configured to support HTTPS.

It looks like this bug was probably introduced when the `ssl` and `verify_ssl` parameters got inverted--when their names changed, the new names were not added to the 'whitelist' of accepted query parameters on line 51 of `testinfra/backend/__init__.py`.

Adding `no_ssl` and `no_verify_ssl` to the list allows these query parameters to propagate to the WinRMBackend.

### DIFF
--- a/testinfra/backend/__init__.py
+++ b/testinfra/backend/__init__.py
@@ -48,7 +48,7 @@ def parse_hostspec(hostspec):
         kw["connection"] = url.scheme
         host = url.netloc
         query = urllib.parse.parse_qs(url.query)
-        for key in ('sudo', 'ssl', 'verify_ssl'):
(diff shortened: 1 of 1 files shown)


robertsigler:
In case there was any doubt--I don't think the CI build failing is related to my changes.. it looks like even master is failing right now: https://travis-ci.org/philpep/testinfra/jobs/415373327

philpep:
Good catch, merged, thanks for fixing this !

robertsigler:
Thanks :)

PR #365, opened 2018-08-31
Fix iptables doc example

gunhu:
host.iptables().get_rules('table', 'CHAIN') is wrong and it should be host.iptables.rules('table', 'CHAIN')


philpep:
Merged, thanks!

PR #373, opened 2018-09-17
Add ip6tables support to iptables module

ianw:
Add basic ip6tables support to iptables module.  Adds a "version"
argument to the existing module which switches it to use the ip6tables
binary.


philpep:
Merged, thanks!

PR #374, opened 2018-09-17
Work around MarkInfo object warnings

ianw:
Silence the warning:

 RemovedInPytest4Warning: MarkInfo objects are deprecated as they
 contain merged marks which are hard to deal with correctly.

by seeing if metafunc.definition is defined, and if so using the
get_closest_marker() function instead.  This should maintain forward
and backwards compatability as required.


philpep:
Merged, thanks!

PR #378, opened 2018-10-02
Improve CommandResult interface

randombenj:
Add a simple `succeeded/failed` interface which allows for simpler assertion in tests.
It is more readable to have tests like:

```python
assert host.run("false").failed
# or
assert host.run("true").succeeded
```
Instead of:

```python
assert host.run("false").rc != 0
# or
assert host.run("true").rc == 0
```

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -446,6 +446,11 @@ def test_sudo_to_root(host):
     assert host.user().name == "user"
 
 
+def test_command_execution(host):
+    assert host.run("false").failed
+    assert host.run("true").succeeded
+
+
 def test_pip_package(host):
     assert host.pip_package.get_packages()['pip']['version'] == '9.0.1'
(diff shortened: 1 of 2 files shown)


philpep:
Hi, thanks for contributing!

We already have `host.check_output(command)` which explicitly check that `rc == 0` before returning stdout.
And we also have `host.run_expect([0], command)` which check return code is in given list and return `CommandResult` object.

But I also like this syntax with `failed / succeeded` so I'd be ok to merge this. I keep MR opened for 1 or 2 days if you or others have others suggestions. Thanks!

randombenj:
I saw the `check_output` and I think it is really nice, but for writeing some simple assertions you sometimes don't want to do exception handling.

Shure, I am open for suggestions, thanks for your input!

philpep:
Merged, thanks!

PR #379, opened 2018-10-16
add support for setting 'become' when running Ansible module

jzafran:
Adds support for setting `become` when running an Ansible module:

```py
host.ansible('service', 'name=docker-compose state=restarted', check=False, become=True)
```

The  `sudo()` context manger does not work for Ansible modules; calls to `host.ansible()` are passed to `AnsibleRunner.run()` without going through `get_command` (which is what [`sudo()` monkey patches `get_sudo_command`](https://github.com/philpep/testinfra/blob/master/testinfra/modules/sudo.py#L48) in as). See `test_ansible_module_become` test for evidence of this.

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -358,6 +358,23 @@ def test_ansible_module(host):
     assert result['stdout'] == 'foo'
 
 
+@pytest.mark.testinfra_hosts("ansible://debian_stretch",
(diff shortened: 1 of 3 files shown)


philpep:
Awesome !
Merged, thanks for this !

PR #382, opened 2018-10-24
Add ability to pass timeouts to the winrm backend

jbmchuck:
Add ability to pass timeouts to the winrm backend

The default read and operation timeouts in pywinrm are quite
low - 30 and 20 respectively.

Windows machines in common configurations frequently will exceed these
timeouts on initial login.  I'm frequently finding the first of a suite
of tests using the winrm backend against newly built machines would
fail but the rest would succeed.  Increasing the read and operation timeout parameters used by pywinrm fixed the issue.

I'm guessing others have encountered this issue - ansible had an
identical issue filed which is fixed by the same method: ansible/ansible#16873

This change allows specifying these parameters as query parameters
in the host spec.


philpep:
Merged, thanks a lot!

PR #396, opened 2019-01-21
Module SaltRunner

jynolen:
This module runs command over salt-master with cmd "salt-run" instead of salt-call


philpep:
Hi, thansk for contributing.
I dont get the point of having a "salt runner" backend since it only works in local, so you can just bypass testinfra and import and use salt runner API within your test. I'm missing something ?

PR #399, opened 2019-01-22
Salt host matching issue

jynolen:
Running:
``` 
>>> d= get_host("salt://*")
>>> d.salt("pkg.version", "python-34")
```
Produced
```
RuntimeError: Error while running pkg.version(['python34']): {'minion-salt-689012b2ca3ff805df1d0cff218e18d2d372db887d1158dfd6dec934b9658cf7': '3.4.9-1.el7', 'master-salt-689012b2ca3ff805df1d0cff218e18d2d372db887d1158dfd6dec934b9658cf7': '3.4.9-1.el7'}. Minion not connected ?
```

### DIFF
--- a/testinfra/backend/salt.py
+++ b/testinfra/backend/salt.py
@@ -48,11 +48,11 @@ class SaltBackend(base.BaseBackend):
         out = self.client.cmd(self.host, func, args or [])
         hosts = SaltBackend.get_hosts(self.host)
         delta = set(hosts) - set(out.keys())
-        if len(delta) > 0:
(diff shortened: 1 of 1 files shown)


philpep:
Hi, thanks for the PR but you're not supposed to use `get_host()` in this case but rather `get_hosts()` which return a list of testinfra `host` objects on which you can call modules.

However the actual behavior need a specific error message I think.

jynolen:
What do you think about this PR should we close it ?

PR #403, opened 2019-02-07
Create a context manager for changing current working directory

bsanders:
Hello,

I have a use-case which doesn't quite work in testinfra as it stands.

host.run() on a local host eventually calls subprocess.run().  That's great, but I noticed it doesn't pass **kwargs.  The fix is trivial, but I've no idea:

* if this implementation of the fix is acceptable
* what it breaks outside my use case.

I added a test, which fails in docker (see below).  I haven't started digging in yet because I thought I'd check first if this whole idea was a non-starter?

```
test/test_modules.py::test_command_execution_in_chdir[docker://debian_stretch] 
[gw1] [ 93%] FAILED test/test_modules.py::test_command_execution_in_chdir[docker://debian_stretch] 

    def test_command_execution_in_chdir(host):
>       assert host.run("pwd", cwd='/tmp').stdout.strip() == '/tmp'
E       AssertionError: assert '/' == '/tmp'
E         - /
E         + /tmp
```

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -468,6 +468,11 @@ def test_command_execution(host):
     assert host.run("true").succeeded
 
 
+def test_command_execution_in_chdir(host):
+    assert host.run("pwd", cwd='/tmp').stdout.strip() == '/tmp'
+    assert host.run("pwd").stdout.strip() != '/tmp'
+
+
 def test_pip_package(host):
     assert host.pip_package.get_packages()['pip']['version'] == '9.0.1'
     pytest = host.pip_package.get_packages(pip_path='/v/bin/pip')['pytest']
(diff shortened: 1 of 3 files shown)


philpep:
Hi, thanks for contributing !

I'm not sure all arguments of subprocess are useful here, but `cwd` definitively is. But the way you've implemented it only works with "local" backend. What about adding a new context manager `cwd` for this, which could enable it for all modules ? For instance:

```python
with host.cwd('/opt/local/app'):
    assert host.file('./myapp').mode == 755
    assert host.run('./myapp --version') == '1.2.3'
```

Under the hood, all remote commands would change directory before running the command.
I think this code will be similar to `host.sudo()` context manager.
What do you think ?

bsanders:
@philpep 

That sounds reasonable.  I'll change the title to reflect that and take a look at an implementation.

This is the file you're suggesting as a model, yes?  `testinfra/modules/sudo.py`

Issue #401, opened 2019-01-25
AnsibleRunnerV2 broken on upstream changes to PlaybookCLI in devel


jctanner:
```
(venv) [jtanner@jtw530 role.1]$ python -c 'from testinfra.utils.ansible_runner import AnsibleRunnerV2; runner = AnsibleRunnerV2()'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/data/workspace.issues/AP-MOLECULE_COLLECTION_INIT/venv/lib/python3.7/site-packages/testinfra-1.16.0-py3.7.egg/testinfra/utils/ansible_runner.py", line 153, in __init__
    self.cli.options = self.cli.base_parser(
AttributeError: 'PlaybookCLI' object has no attribute 'base_parser'
(venv) [jtanner@jtw530 role.1]$
```

Caused by https://github.com/ansible/ansible/commit/7e92ff823e2bf97ee33c1b7ff87bb8acecb5ae77#diff-b54c875fc6bfd4d98dcc488f06a07bea

singuliere:
The commit introducing the change, from the PR and before the squash post merge: https://github.com/ansible/ansible/pull/50069/commits/39e4421af26584863eb69d8ac01d44c66676c290

cboylan:
Using https://github.com/ansible/ansible-runner may be more reliable long term method of running ansible from testinfra. I don't have enough knowledge around testinfra or ansible-runner to know for sure, but wanted to make sure this alternative was called out since ansible-runner seems built for this exact use case.

jctanner:
I've already started a port to runner and it's only held up right now by an obscure bug: https://github.com/ansible/ansible-runner/issues/200

ianw:
@jctanner thanks for that; I took that gist and got it working for at least a loopback situation and will test more.  I missing what was happening with the event capture, ascii codes and molecule, but I hope that with you as the author it can be fixed.  OpenStack should be able to test that pull request as it progresses in our CI too, as an extra level of testing over travis

philpep:
Great ! +1 for using ansible-runner

decentral1se:
:rocket: :rocket: :rocket: 

`ansible-runner` integration would indeed be fantastic.

philpep:
Hi, I just released 3.0.0 with the new ansible runner which should be compatible with ansible 2.8 please sse the changelog here: https://testinfra.readthedocs.io/en/latest/changelog.html

PR #404, opened 2019-02-08
[LOOKING FOR SOMEONE TO TAKEOVER :-] replace ansible base_parser with create_base_parser

singuliere:
As of https://github.com/ansible/ansible/pull/50069 the base_parser
function no longer exists. Use
ansible.arguments.optparse_helpers.create_base_parser
instead.

Fixes: https://github.com/philpep/testinfra/issues/401

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ services:
   - docker
 env:
   matrix:
-    - TOXENV=py27,ansible1,py35
+    - TOXENV=py{27,35}-ansible{2,devel},ansible1
     - TOXENV=flake8
     - TOXENV=pylint
     - TOXENV=sphinxdoc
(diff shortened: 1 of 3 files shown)


singuliere:
This is only a draft and turns out to more involved than a function name replaced with another. The argument parsing and storage moved from being methods of the CLI class and derived classes to partly static methods and in the end the resulting options is stored in a global variable instead of a data member.

singuliere:
@philpep this is above my paygrade and I would very much appreciate your guidance on how to proceed with this change. Should I keep going in this direction or would you recommend another approach?

decentral1se:
Looks like https://github.com/philpep/testinfra/pull/410 is another direction for a solution!

singuliere:
@decentral1se indeed! Let's close this one then.

PR #410, opened 2019-03-01
Switch to ansible-runner

ianw:
Switch to ansible-runner instead of directly interfacing with ansible

Based on @jctanner gist at
https://gist.github.com/jctanner/512ec07171d95e61a5b7fd5f9103a301

I hope we can collaborate around something like this.  I'm making a common pull request but please update ... our infra can test pull requests from the main repo

Not ready to merge, obviously, but this reliably works for the backend tests (and my local tests) with both local and remote ansible that go via the ```shell``` module.

The two tests that fail are not using ``shell`` output.  As mentioned inline; I can't seem to get the json response from ansible-runner reliably.  I get the feeling ansible-runner is sometimes dropping some of the output of the ansible process in it's ``stdout`` descriptor it returns.  This is a problem for the things *not* calling the ```shell``` module.

### DIFF
--- a/testinfra/utils/ansible_runner.py
+++ b/testinfra/utils/ansible_runner.py
@@ -79,7 +79,7 @@ class AnsibleRunnerV2(object):
             cmd.append('--list')
 
         try:
-            so = subprocess.check_output(cmd)
+            so = subprocess.check_output(cmd, universal_newlines=True)
         except subprocess.CalledProcessError as e:
             raise AnsibleInventoryException(e)
 
--- a/testinfra/utils/ansible_runner.py
+++ b/testinfra/utils/ansible_runner.py
@@ -182,4 +182,3 @@ class AnsibleRunner(object):
             return significant_event
 
         return {}
-
--- a/testinfra/utils/ansible_runner.py
+++ b/testinfra/utils/ansible_runner.py
@@ -34,12 +34,7 @@ if six.PY2:
 else:
     import tempfile
 
-__all__ = ['AnsibleRunner', 'to_bytes']
-
-
-def to_bytes(data):
-    '''Why!?!?'''
-    return b'%s' % data
+__all__ = ['AnsibleRunner']
 
 
 class AnsibleInventoryException(Exception):
--- a/testinfra/utils/ansible_runner.py
+++ b/testinfra/utils/ansible_runner.py
@@ -73,28 +73,25 @@ class AnsibleRunnerV2(object):
     def fetch_inventory(self, host=None):
         '''Helper function for ansible-inventory'''
 
-        cmd = 'ansible-inventory -i %s' % self.host_list
+        cmd = [
+            'ansible-inventory',
+            '-i',
+            self.host_list
+        ]
         if host is not None:
-            cmd += ' --host=%s' % host
+            cmd.append('--host=%s' % host)
         else:
-            cmd += ' --list'
+            cmd.append('--list')
 
-        env = os.environ.copy()
-        env['ANSIBLE_NOCOLOR'] = "1"
-        p = subprocess.Popen(
-            cmd,
-            shell=True,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            env=env
-        )
-        (so, se) = p.communicate()
+        try:
+            so = subprocess.check_output(cmd)
+        except subprocess.CalledProcessError as e:
+            raise AnsibleInventoryException(e)
 
-        if p.returncode != 0:
-            msg = 'ansible-inventory failed: %s' % se
-            raise AnsibleInventoryException(msg)
-
-        inv = json.loads(so)
+        try:
+            inv = json.loads(so)
+        except json.JSONDecodeError as e:
+            raise AnsibleInventoryException(e)
 
         return inv
 
--- a/testinfra/utils/ansible_runner.py
+++ b/testinfra/utils/ansible_runner.py
@@ -18,6 +18,7 @@
 from __future__ import unicode_literals
 from __future__ import absolute_import
 
+import fnmatch
 import json
 import os
 import shutil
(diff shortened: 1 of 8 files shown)


testinfra/utils/ansible_runner.py
```diff
@@ -18,150 +18,189 @@
...
+            stdout = "\n".join(final_stdout).strip()
+
+            result = {
+                'rc': r.rc,
+                'stdout': stdout,
+                'stderr': None
+            }
+            return result
+        else:
+            return json.load(r.stdout)
```
ianw: This seems to be pretty close, except ```r.stdout``` doesn't actually seem to consistently contain the full stdout from ansible.  This is especially evident with

```
setup = host.ansible("setup")["ansible_facts"]
````

which returns a very big json list ... it's often cut off.  I think there's something wrong with the chunking done by ansible-runner in the way it copies into stdout.  thus this load just fails on invalid json

File: testinfra/utils/ansible_runner.py [hunk 2]
```diff
@@ -18,150 +18,189 @@
...
+        # newlines from something like:
+        #
+        # ---
+        # localhost | CHANGED | rc=0 >>
+        # Linux
+        #
+        # ---
+        #
+        # I am sure this is not the best way...
+        if module_name == "shell":
```
ianw: When using the shell module, the output is expected to just be the stdout of whatever was called; strip all the ansible output.  Otherwise, it should return the json output as an object

philpep:
Hi, thanks for trying this. I'm glad to see ansible starting writting some "user-friendly" python API because maintaining this in testinfra is very costly (however I don't understand why ansible-runner is not included in ansible...)

Looking forward to see this PR ready to merge.

jctanner:
per https://github.com/ansible/ansible-runner/issues/200 it appears that molecule's usage of colorama is causing issues for runner.

decentral1se:
Molecule now provides a PY_COLORS toggle (see https://github.com/ansible/molecule/pull/1533) :+1:

matburt:
What does this mean for the runner integration?

ianw:
This change is based on https://github.com/philpep/testinfra/pull/428 and now passes the test-suite locally for me.  I think we should squash most of this into a single, understandable change, but I'm not sure how other people like to work ... I'm happy to do that.  A couple of the changes are separate pull requests because they are things I found that stand alone.

ianw:
There seems to be something up with pip/virtualenv installing in travisci right now

I have tested this in https://review.opendev.org/#/c/656877/

The interesting bit is http://logs.openstack.org/77/656877/5/check/system-config-run-base-ansible-devel/a50c4ad/job-output.txt.gz#_2019-05-03_12_02_45_978799

For reference, here is a run using the status quo http://logs.openstack.org/90/651390/4/check/system-config-run-base/7375ffb/job-output.txt.gz#_2019-05-02_17_19_48_306100

I think we know this is slower.  Looking at the runtimes for the same 58 tests we seem to have a base line:

```
2019-05-02 17:19:49.422402
2019-05-02 17:20:41.476163 
(~1 minute)
```

```
2019-05-03 12:02:45.984555 
2019-05-03 12:09:41.150059 
(~7 minute)
```

philpep:
@ianw thanks for the tests !

Yes it's indeed very slower than before, that's why I propose to use testinfra backend instead (except for the "ansible" module) in #431

Can you consider making changes to use ansible-runner on top of my series in #432 ?

I see you implemented https://github.com/philpep/testinfra/pull/410/commits/4f00901ea3f115f05cc14caf2a1b0a861f89836a that I also have in my series https://github.com/philpep/testinfra/pull/432/commits/6ff3c409e6d3145afe576fa1c169647dabd32460#diff-2c2050402fbd38935790464dc2668909R109 I wonder how do they compare ?

In particular I added some useful tests on my series, so it might be worth using it as a base, ansible-runner has just to be implemented in `run_module()` instead of `ansible --tree` call.

But I'm still not convinced we need to write this `envvars` file and handle the private_data_dir here.

philpep:
Well to be honest, I'm still not conviced by ansible-runner at all for now. This add a dependency/wrapper that doesn't provide any additionnal feature to us over calling `ansible --tree` which just "works". Hopefully ansible-runner will provide a performant ansible connection tunnel through the target host, then I'll have another opinion about this.

ianw:
> I see you implemented [4f00901](https://github.com/philpep/testinfra/commit/4f00901ea3f115f05cc14caf2a1b0a861f89836a) that I also have in my series [6ff3c40#diff-2c2050402fbd38935790464dc2668909R109](https://github.com/philpep/testinfra/commit/6ff3c409e6d3145afe576fa1c169647dabd32460#diff-2c2050402fbd38935790464dc2668909R109) I wonder how do they compare ?

Yes, after I did that, I realised you must have done the same because it was using ansible-inventory!  But i'm glad we came out with something that looks similar :)

I think that children: can be nested indefinitely, so it needs to be recursive?

I also got a bit confused as to what "pattern" is -- is it a string or a list?  You can see at https://opendev.org/opendev/system-config/src/branch/master/testinfra/test_adns.py#L16 an example of how we use it in our CI, we are pretty much tagging each test_* file with ``testinfra_hosts = [list-of-hosts]``.  I see it come in as both?

What is interesting though, and I think it might be a bug, is that if there are *no* hosts matching, we get a test parameterized as "_testinfra_host0" in the list, which just gets skipped.  We split our jobs up, so for example we're only running all the testing on DNS servers if someone actually changes the ansible roles that affect DNS servers.  For example, in our base test (the roles that happen on *every* host) ```get_hosts()``` will return an empty list for the DNS test I referenced above (```testinfra_hosts = ['adns1.opendev.org']```) because we don't have ```adns1.opendev.org``` in the inventory.

```
2019-05-03 12:02:45.978799 | bridge.openstack.org | collecting ... collected 58 items
2019-05-03 12:02:45.981093 | bridge.openstack.org |
2019-05-03 12:02:45.984555 | bridge.openstack.org | testinfra/test_adns.py::test_bind[_testinfra_host0] SKIPPED              [  1%]
2019-05-03 12:02:45.986828 | bridge.openstack.org | testinfra/test_adns.py::test_zone_files[_testinfra_host0] SKIPPED        [  3%]
```

It seems it would be better for the test just to not be there.  Anyway, that is something we can look at separately.

ianw:
> Can you consider making changes to use ansible-runner on top of my series in #432 ?

Yes this seems reasonable way forward; I can look into this next week.  We can also do similar testing of #432 (feel free to use opendev infra testing at any time, based on https://review.opendev.org/#/c/656877/ ... all needs to happen is switch the Depends-On: line to test different pull requests.  You'll just need a gerrit account and can either start a separate review or feel free to edit that one ... anyone in #openstack-infra freenode will by more than happy to help :)

I agree the performance is probably a blocker ATM.  I hope that having this so close can get people interested enough to overcome this and the we have a nice standardisation across some big bits of the ecosystem.

philpep:
Superseeded by  #432

Issue #330, opened 2018-05-31
winrm limitations


potchin:
Not sure if this is a bug or something which has not yet been implemented yet. If the latter it should probably be documented.

I just tried to call a `host.file(filename).exists` against a windows machine and it failed when it tried to call `uname` which will obviously fail..
```
INFO:testinfra:RUN CommandResult(command=u'uname -s', exit_status=1, stdout=None, stderr="'uname' is not recognized as an internal or external command,\r\noperable program or batch file.\r\n")
```

Which methods are supported when using the winrm backend?

philpep:
Hello, Indeed, for now the support on windows is limited to running commands with `run()`, `check_output()`.

I think `file` module could work if someone write a windows compatible module.

PR #425, opened 2019-04-13
Add Windows support

trevor-viljoen:
This adds some basic support for Windows packages (chocolatey) and users. I'm not sure how to add tests for this because the Windows docker images that are available require a Windows host to run.  Also, I apologize for this in advance. We needed to be able to test some things at work, so I thought I'd see if I could add some very basic Windows support. This helps to start addressing #330, but I haven't yet worked out a solution for the file module.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -10,6 +10,7 @@ disable=
     inconsistent-return-statements,
     keyword-arg-before-vararg,
     useless-object-inheritance,  # to be removed for py3
+    duplicate-code,  # too many false positives between file and user modules :/
     wrong-import-order
 max-branches=20
 max-bool-expr=6
--- a/testinfra/modules/package.py
+++ b/testinfra/modules/package.py
@@ -65,6 +65,8 @@ class Package(Module):
 
     @classmethod
     def get_module_class(cls, host):
+        if host.system_info.type == 'windows':
+            return ChocolateyPackage
         if host.system_info.type == "freebsd":
             return FreeBSDPackage
         if host.system_info.type in ("openbsd", "netbsd"):
@@ -189,6 +191,7 @@ class ArchPackage(Package):
     @property
     def release(self):
         raise NotImplementedError
+
 
 class ChocolateyPackage(Package):
 
--- a/testinfra/modules/package.py
+++ b/testinfra/modules/package.py
@@ -64,7 +64,7 @@ class Package(Module):
         return "<package %s>" % (self.name,)
 
     @classmethod
-    def get_module_class(cls, host):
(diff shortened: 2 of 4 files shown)


philpep:
Hi @trevor-viljoen , thanks for the patches !
I added some commit with minor changes, can you tell me if it still works for you ?
(especially I'm not sure about exit code for "command not found").

Then I'll be happy to merge this !

trevor-viljoen:
I’m on holiday, but will try to check this out as soon as I can. From looking at the changes you made, I don’t see anything that should break anything. I’ll check on the error code.

philpep:
Ok thanks!
Especially I'm not sure about exit code 127 for "command not found", otherwise the PR should work just like before my amends.

trevor-viljoen:
@philpep Running the tests I wrote with the changes above, I get this error:
```
E       AssertionError: Unexpected exit code 1 for CommandResult(command='uname -s', exit_status=1, stdout=None, stderr=b"'uname' is not recognized as an internal or external command,\r\noperable program or batch file.\r\n")
```

philpep:
@trevor-viljoen thanks, I updated the code accordingly. I think we can do better to detect windows but I'm ok to move forward with the current code.

trevor-viljoen:
Looks good to me. One idea to better detect Windows would be to check for `stderr`'s output. `'uname' is not recognized as an internal or external command, operable program or batch file.`

```python
if uname.rc == 1 and "batch file" in uname.stderr:
    sysinfo.update(**self._get_windows_sysinfo())
    return sysinfo
```

trevor-viljoen:
This works for me.

```python
from packaging import version as pkgver


@pytest.mark.parametrize("name, version", [
    ("vcredist2010", "10.0.40219.2"),
])

def test_packages(host, name, version):
      """ Make sure packages are installed and correct version """
      pkg = host.package(name)
      assert pkg.is_installed
      pkg_version = pkgver.parse(pkg.version)
      assert pkg_version > pkgver.Version("10.0")

def test_user(host):
      """ Make sure vagrant user exists """
      user = host.user('vagrant')
      assert user.exists
      assert user.name == 'vagrant'
      assert 'Administrators' in user.groups
      assert 'Users' in user.groups
```

philpep:
> Looks good to me. One idea to better detect Windows would be to check for `stderr`'s output. `'uname' is not recognized as an internal or external command, operable program or batch file.`

Yes but I'm scared the message could be internationalized with the user language here :)

The PR looks good to me, I'll merge as soon as the travis build is green.
I also opened https://github.com/philpep/testinfra/issues/429 If you have some time to spend on windows testing.

Thanks!

Issue #431, opened 2019-04-29
Ansible runner issues


philpep:
Here I sumarize issues related to the ansible backend runner.

Ansible has been available as a "connection backend", meaning the use of ansible python api to execute commands on remote hosts. This have several issues:

* The code is very hard to maintain because ansible doesn't have a stable python API. From https://docs.ansible.com/ansible/latest/dev_guide/developing_api.html "This API is intended for internal Ansible use. Ansible may make changes to this API at any time that could break backward compatibility with older versions of the API. Because of this, external use is not supported by Ansible." We are currently no longer compatible with ansible 2.8

* We are actually violating the license because ansible is GPL and testinfra is not (apache2), so importing ansible in python require the code to be under GPL.

Requirements:

* I want to exclude all solutions including writting code with the ansible python API. Every new release of ansible produce a large amount of work and frustration to me.

* Fully backward compatible, at least for general usage (molecule should not break)

* Should be fast (at least no slower than now).

Possible solutions:

https://github.com/ansible/ansible-runner first appeared to be a possible solution, but after reading the documentation and a bit of code, I think it's actually not providing the feature we need (note that I might have missed something...)

I considered interfacing with ansible cli directly, but this will introduce a terrible performance regression.

I think people using testinfra and ansible use the ansible backend for "convenience", because it can read the inventory, but only some tests actually require ansible (only those using the ansible module). Also I think most people use ansible with either local, ssh or docker connection and testinfra already have such connection backends.

So I propose to:

* Write some code to parse/decode ansible configuration and inventory and convert them to testinfra local, ssh or docker backend. ``ansible-config`` and ``ansible-inventory`` tools can help here.

* Reimplement the ansible module using ansible cli called in subprocess.

This will drop support for other ansible connection plugins (https://docs.ansible.com/ansible/latest/plugins/connection.html), but I think we can implement them later, we already have some code for kubectl, lxc, winrm.

I already have an POC implementation of this and it works pretty well. I'll submit a PR soon.

Please leave a comment if you want to point me on your use case, I'll be happy to test implementation with your code !

Also special thanks to people that have tried to fix this issue by submitting PRs, especially @jctanner and @ianw , your work helped me a lot for sumarizing this issue and find this solution !

Related to #428 #410 and #401

jctanner:
@philpep fair enough, but i'm still curious if you could elaborate further on what is missing from runner.

PR #432, opened 2019-04-29
New ansible runner

philpep:
Implementation for https://github.com/philpep/testinfra/issues/431

### DIFF
--- a/testinfra/utils/ansible_runner.py
+++ b/testinfra/utils/ansible_runner.py
@@ -108,15 +108,24 @@ class AnsibleRunner(object):
     def get_hosts(self, pattern=None):
         inventory = self.inventory
         result = set()
+
+        def itergroup(group):
+            for host in inventory[group].get('hosts', []):
+                yield host
+            for g in inventory[group].get('children', []):
+                for host in itergroup(g):
+                    yield host
+
         if inventory == EMPTY_INVENTORY:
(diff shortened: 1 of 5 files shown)


testinfra/utils/ansible_runner.py
```diff
@@ -11,56 +11,177 @@
...
+            self._host_cache[host] = get_ansible_host(
+                self.ansible_config, self.inventory, host)
+            return self._host_cache[host]
+
+    def run(self, host, command):
+        return self.get_host(host).run(command)
+
+    def run_module(self, host, module_name, module_args, become=False,
+                   check=True, **kwargs):
+        cmd, args = 'ansible --tree %s', [None]
```
jctanner: I had no idea this existed. Pretty neat!

decentral1se:
Sorry, kinda out of the loop but what is the relation of this PR with https://github.com/philpep/testinfra/pull/410? In the case of manual testing this new runner with testinfra under molecule, I should be aiming to try this PR branch? Thanks for the good work!

philpep:
I merge this PR. I think it's a good enough base to work with. It should be fully backward compatible, enough performant and fixes all the issues from #431 . It think we will probably have small regressions , I bet around sudo or with some complex usages in test suites, so this will go in a major release 3.0.0

PR #433, opened 2019-05-02
Add container env to debian Stretch Dockerfile

ianw:
systemd requires the "container" variable be set when it is running
inside a container [1].  Without this, on some hosts the container can
go into a silent purgatory enforced by selinux where systemd is
running, but sshd or other services never start in the container
... leading to ansible tests simply hanging forever trying to ssh in.

[1] https://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/

### DIFF
--- a/images/debian_stretch/Dockerfile
+++ b/images/debian_stretch/Dockerfile
@@ -1,5 +1,7 @@
 FROM debian:stretch
 
+ENV container docker
+
 # ntpd is linked to /bin/false, so the service is not running but is enabled
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get -y install --no-install-recommends \


ianw:
@philpep this appears to have failed travis CI in alpine looking for sshd -- I don't think that's related.  However, alpine is the container that seems to take over my fedora host somehow ...

```
______________________ test_ssh_service[docker://alpine] _______________________
[gw0] linux -- Python 3.5.6 /home/travis/build/philpep/testinfra/.tox/py3/bin/python
host = <testinfra.host.Host object at 0x7f835bbb7208>, docker_image = 'alpine'
    @all_images
    def test_ssh_service(host, docker_image):
        if docker_image in ("centos_6", "centos_7",
                            "alpine", "archlinux"):
            name = "sshd"
        else:
            name = "ssh"
    
        ssh = host.service(name)
        if docker_image == "ubuntu_xenial":
            assert not ssh.is_running
        # FIXME: is_running test is broken for archlinux for unknown reason
        elif docker_image != "archlinux":
>           assert ssh.is_running
E           assert False
E            +  where False = <service sshd>.is_running
```

philpep:
Merged, thanks for digging into this issue !
Yes, alpine sshd is a known flapping test (it fail when ssh is "starting" but not "started"), It might be the same issue than archlinux referenced by the "FIXME".

PR #436, opened 2019-05-06
Handle literal ipv6 addresses

ianw:
An ansible inventory might specify the "ansible_host" as a literal
ipv6 address.

This needs handling in ingestion; in ansible_runner.py we add a check
if the host looks like a literal ipv6 address and enclose it in []'s
per RFC2732 (a util function is added for this, which works better on
Python 3).

When connecting, testinfra's host parsing in backend/base.py then
needs to check for []'s in the connection string and extract the
literal ipv6 address if found.  The test-case for this is updated with
a few ipv6 examples.

### DIFF
--- a/test/test_backends.py
+++ b/test/test_backends.py
@@ -233,6 +233,14 @@ def test_docker_encoding(host):
     ('pr%C3%A9nom@h', HostSpec('h', None, 'prénom', None)),
     ('pr%C3%A9nom:p%40ss%3Aw0rd@h', HostSpec('h', None, 'prénom',
                                              'p@ss:w0rd')),
+    # ipv6 matching
+    ('[2001:db8:a0b:12f0::1]',
+     HostSpec('2001:db8:a0b:12f0::1', None, None, None)),
+    ('user:password@[2001:db8:a0b:12f0::1]',
+     HostSpec('2001:db8:a0b:12f0::1', None, 'user', 'password')),
+    ('user:password@[2001:4800:7819:103:be76:4eff:fe04:9229]:22',
+     HostSpec('2001:4800:7819:103:be76:4eff:fe04:9229', '22',
+              'user', 'password')),
 ])
 def test_parse_hostspec(hostspec, expected):
     assert BaseBackend.parse_hostspec(hostspec) == expected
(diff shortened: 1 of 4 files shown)


ianw:
An example of the failure is at http://logs.openstack.org/77/656877/6/check/system-config-run-base-ansible-devel/e20d459/job-output.txt.gz#_2019-05-06_00_28_42_477484

The inventory being used there is http://logs.openstack.org/77/656877/6/check/system-config-run-base-ansible-devel/e20d459/bridge.openstack.org/etc/ansible/hosts/inventory.yaml

I will endeavour to get this checked in our CI

@philpep could we discussing including the opendev CI to comment in a similar way Travis-ci does?  Our test is fairly good (if I do say so ;) ... we setup 4 different hosts with 4 different OS's, run our playbooks against them and then ensure the playbooks worked with testinfra.  There's a range of environments from ipv4 only, mixed and ipv6 only, depending on which provider the tests runs on.  The comment would look like https://github.com/ansible/ansible/pull/54532#issuecomment-477666909.  We can certainly do things like filter on what file paths to run this check on (e.g., to only run it if it touches backend/ansible related files, for example), if you'd like.  I think this would be mutually beneficial to avoid breakages.

ianw:
The travisci failure here seems to be related to a docker timeout

I have tested in opendev CI and it has passed

http://logs.openstack.org/77/656877/7/check/system-config-run-base-ansible-devel/de9043d/job-output.txt.gz#_2019-05-06_02_47_22_597322

This was an ipv6 environment as specified by http://logs.openstack.org/77/656877/6/check/system-config-run-base-ansible-devel/e20d459/bridge.openstack.org/etc/ansible/hosts/inventory.yaml

philpep:
Merged, thanks!

Yeah i'm totally open to add opendev CI here ! How does the integration with github works ?

ianw:
@philpep thanks, it's just a matter of adding it to our configurations:

https://review.opendev.org/657461 Add testinfra 3rd party testing 

If you'd like to get an account and give a positive vote on that, it would be great.  We can also put in filters (see files: section, which is used e.g. https://opendev.org/openstack/project-config/src/branch/master/zuul.d/projects.yaml#L37 if you'd prefer to run only a subset of pull requests that touch certain things

PR #441, opened 2019-05-08
Fix sudo for paramiko backend

urtow:
cfg does not contain RequestTTY item, so with lines sudo can`t get tty


philpep:
Merged, thanks!

PR #448, opened 2019-05-28
Support `ansible_private_key_file` as well

decentral1se:
Hi there, we received a report on https://github.com/ansible/molecule/issues/2069 about Molecule failing to make an SSH connection and after a quick glance over in https://github.com/ansible/molecule/issues/2069#issuecomment-496456638, I think I have a patch.

The short of it is, Ansible advertises that you can use both "ansible_ssh_private_key_file" and "ansible_private_key_file" for specifying SSH connection options and Molecule happens to use the "ansible_private_key_file" by default. The new Ansible runner implementation here in Testinfra (from what I can tell) does not pick this up.

Didn't dig too deeply on this but hopefully this goes in the right direction ...

### DIFF
--- a/test/test_backends.py
+++ b/test/test_backends.py
@@ -161,6 +161,14 @@ def test_ansible_get_variables():
         'host.port': '2222',
         'ssh_identity_file': 'key',
     }),
+    ('host', {}, b'host ansible_host=127.0.1.1 ansible_user=u ansible_private_key_file=key ansible_port=2222 ansible_become=yes ansible_become_user=u', {  # noqa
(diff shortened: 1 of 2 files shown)


decentral1se:
CI failure appears to be unrelated.

philpep:
Looks reasonable to handle both, Merged, thanks!

decentral1se:
Thank you very much. I must now obligatorily beg you for a new release :)

philpep:
Sure, It's in 3.0.5 ;)

decentral1se:
<3

Issue #452, opened 2019-05-31
Paramiko connections don't get closed


krepflap:
Running my tests against a large inventory, I can see I'm hitting the max open file descriptors with 1000+ `ESTABLISHED` connections belonging to py.test process.

Can you see how this behaves at your end? Probably we need to close the connections during testing? Thanks :)

http://docs.paramiko.org/en/2.4/api/client.html#paramiko.client.SSHClient.close

philpep:
Hi, thanks for reporting !

How many hosts do you have in your inventory ? It shouldn't be more that 1 connection per host.

krepflap:
Hi, thanks for your reply.

I'm using a large ansible inventory with 3000+ hosts. I'm still in testing/proof of concept phase to see if testinfra is a good fit for that kind of inventory size, but I think it should be possible. Anyway, after a while I reach the limits of the system and load goes up because of all the connections that remain open (limited by the default "open files" limit of 1024 of the system I believe).

I'm not sure how ansible handles this internally.

philpep:
Ok, I think ansible doesn't have this issue because it use plain ssh connection instead of paramiko with options to keep the connection open for 1 minute: https://docs.ansible.com/ansible/2.8/reference_appendices/config.html#ansible-ssh-args

I think we should do the same in testinfra.

philpep:
Also I think we will have issue with test ordering, since pytest will run all tests one by one for each host, it should be more efficient to test host by host.

krepflap:
Thanks for replying.

I'm getting by with paramiko and a patched version for now. It's not so pretty and of course a bit slower, but I think it allows me to do what I need (coupled with xdist).

https://github.com/krepflap/testinfra/commit/069737baf5ae076552541e1ead9c2b7166eb1c33

I think for large inventories there is room for optimization as you say, but it's already quite usable. I was doing this sort of testing with ansible before ( based on https://blog.newrelic.com/engineering/ansible-auditing-tool/ ) and that was definitely much more finicky.

PR #454, opened 2019-06-05
Add extras for backend dependencies

ianw:
This adds environment selectors for backends in setup.cfg.  The
inspiration here is that the ansible backend actually requires
paramiko but it isn't specified anywhere.  Having environment markers
makes it a bit eaiser in CI environments to ensure everything is
present.

I've taken a guess at the requirements for the others; although
they're simple I think it's good for consistency to have all backends
with their own environment.  A documentation note is added too.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,3 +3,14 @@ universal = 1
 
 [tool:pytest]
 norecursedirs = .tox .git .local *.egg build
+
+[options.extras_require]
+ansible = ansible>=2.8,<3; paramiko
+docker =
+kubectl =
+local =
+lxc =
+paramiko = paramiko
+salt = salt; tornado
+winrm = pywinrm
+
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,12 +5,11 @@ universal = 1
 norecursedirs = .tox .git .local *.egg build
 
 [options.extras_require]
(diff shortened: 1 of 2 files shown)


ianw:
Related upstream review: https://review.opendev.org/#/c/662662

philpep:
Hi, thanks for this !

Why hard dependency on ansible>=2.8 ? The code actually run fine with 2.7, potentially with earlier versions too. I think >=2.4 should be suitable here.
Also paramiko isn't a hard dependency for testinfra+ansible, we also handle docker here.
Furthermore I plan to use ssh instead of paramiko as default backend for ansible hosts, because ssh is the default connection in ansible and also because of #452

ianw:
recheck

setup.cfg
```diff
@@ -3,3 +3,14 @@ universal = 1
...
 [tool:pytest]
 norecursedirs = .tox .git .local *.egg build
+
+[options.extras_require]
+ansible = ansible>=2.8,<3; paramiko
+docker =
+kubectl =
+local =
+lxc =
+paramiko = paramiko
```
arthurlutz: I would find this usefull ;-)

philpep:
Merged with minor modifications, thanks!

PR #464, opened 2019-07-23
get variables from ansible playbook

w1ndblow:
added parameter --ansible-playbook. for example --ansible-playbook=test-playbook.yml
the process of applying variables in "vars" in the playbook is:
1) applying variables from play with "all" value of "hosts" directive
2) applying variables from play with hostname value of "hosts" directive


philpep:
Hi, this looks interesting, thanks for this.

However I think i'd be preferable to pass the playbook file in `ansible.get_variables(playbook="/path/to/playbook.yml")` (from the ansible module). Rather than from the command line. This would allow to read variables from any playbook.

w1ndblow:
I think, both sources of "playbook" can stay. What do you think? If you insist on one variant, I'll do that.

PR #466, opened 2019-07-26
DEBUG log when running commands

hfuss:
For those using Ansible inventories and who want to log in their `pytest`'s with `INFO` logging, this currently will log sensitive vault variables. It'd be preferable if this was just a `DEBUG` log instead.

This is the only `INFO` log I encounter in `testinfra` when using an `ansible` connection so hopefully this won't upset anyone. Thanks for a great plugin!

### DIFF
--- a/testinfra/backend/base.py
+++ b/testinfra/backend/base.py
@@ -278,5 +278,5 @@ class BaseBackend(object):
 
     def result(self, *args, **kwargs):
         result = CommandResult(self, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


hfuss:
Example of current `INFO` log with `stdout` redacted
```
INFO     testinfra:base.py:281 RUN CommandResult(command='ansible-inventory --list -i operations/inventories/temp', exit_status=0, stdout=*****
```

philpep:
This LGTM, merged, thanks!

Issue #470, opened 2019-08-02
get_ansible_inventory: ValueError: No JSON object could be decoded


ssbarnea:
Location of issue https://github.com/philpep/testinfra/blob/master/testinfra/utils/ansible_runner.py#L58

In order to unblock current users we need to implement a workaround in testinfra: disabling ansible verbosity for this command via `ANSIBLE_VERBOSITY=0`.

Upstream bug: https://github.com/ansible/ansible/issues/59973
Downstram bug: https://github.com/ansible/molecule/issues/2188

ssbarnea:
@philpep @ianw Can you please check the the two proposed PR for fixing this bug? Any would them would be ok, I do not care much which but I need a new patch release for molecule.

decentral1se:
Solid detective work. I prefer https://github.com/philpep/testinfra/pull/471 and have left a review.

PR #471, opened 2019-08-02
Avoid JSON parsing error in get_ansible_inventory

ssbarnea:
Avoids Ansible bug https://github.com/ansible/ansible/issues/59973 which
prevents testinfra from parsing inventory by disabling ansible verbosity.

Fixes: #470

PR #472, opened 2019-08-02
Avoid JSON parsing error in get_ansible_inventory (v2)

ssbarnea:
Avoids Ansible bug https://github.com/ansible/ansible/issues/59973
which prevents testinfra from parsing inventory by using
named file output instead of stdout.

Fixes: #470


philpep:
Closing in flavor of #471 (``--output`` only works for ansible >= 2.8)

Issue #411, opened 2019-03-06
run_local not working on windows


jacobEAdamson:
My system is Python 3.7.0 and I receive an error when I try to run anything with run_local on windows (which affects all backends). The stack trace looks like this:

```text
venv\lib\site-packages\testinfra\host.py:71: in run
    return self.backend.run(command, *args, **kwargs)
venv\lib\site-packages\testinfra\backend\docker.py:35: in run
    "docker exec %s /bin/sh -c %s", self.name, cmd)
venv\lib\site-packages\testinfra\backend\base.py:203: in run_local
    stderr=subprocess.PIPE,
..\..\..\appdata\local\programs\python\python37-32\Lib\subprocess.py:756: in __init__
    restore_signals, start_new_session)
..\..\..\appdata\local\programs\python\python37-32\Lib\subprocess.py:1100: in _execute_child
    args = list2cmdline(args)
..\..\..\appdata\local\programs\python\python37-32\Lib\subprocess.py:511: TypeError
    TypeError: argument of type 'int' is not iterable
```

Looks like the fact that the command is being encoded before running is to blame. Popen doesn't handle byte arrays the same way it does on windows. Was able to fix by modifying code to this:

```python
def run_local(self, command, *args):
    command = self.quote(command, *args)
    if os.name != 'nt':
        command = self.encode(command)
    p = subprocess.Popen(
        command, shell=True,
        stdin=subprocess.PIPE,
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
    )
    stdout, stderr = p.communicate()
    result = self.result(p.returncode, command, stdout, stderr)
    return result
```

Basically only encode when not on windows

ianw:
Thanks, from @jacobEAdamson comments in the pull request and seeing here you mention python 3.7 things are a little more clear, and a little more confused.

```self.encode()``` is going to return a bytes object on python3 which will mean the following 

```
$ python
Python 2.7.16 (default, Apr 30 2019, 15:54:43) 
>>> subprocess.list2cmdline(b"testing 123")
't e s t i n g " " 1 2 3'

$ python3
Python 3.7.4 (default, Jul  9 2019, 16:32:37) 
>>> subprocess.list2cmdline(b'testing 123')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.7/subprocess.py", line 530, in list2cmdline
    needquote = (" " in arg) or ("\t" in arg) or not arg
[...]

PR #480, opened 2019-08-13
#411 - Apply @jacobEAdamson's suggested changes

theothertom:


testinfra/backend/base.py
```diff
@@ -193,7 +194,8 @@ def run(self, command, *args, **kwargs):
 
     def run_local(self, command, *args):
         command = self.quote(command, *args)
-        command = self.encode(command)
+        if os.name != 'nt':
+            command = self.encode(command)
```
ianw: I'd certainly be willing to believe that there's encoding issues on different platforms.

But the original error seems to be

```
TypeError: argument of type 'int' is not iterable
```

I can not see how a priori how what seems like an "integer has crept in that should be a string" issue relates to string encoding.   So this isn't an obvious merge.

That said, because encoding issues on commandline args for Windows is something I imagine nobody is super conversant in, this change should come with even more explanation than usual in the change, and probably even an inline comment.  Please if you're putting things in for maintainers to look at, add enough context in descriptions for it to be actionable.

theothertom: Hi,

Sorry - probably should have added a bit more context, was mostly just creating a PR for the changes proposed in someone else's issue.

I've added a comment explaining that it's an internal thing inside subprocess to the code. A fuller explanation is:
subprocess.Popen() accepts either "a sequence of program arguments or else a single string" as a command to run.
On POSIX, a string means "the exact command, with no arguments" whereas a byte sequence means "the command and arguments" - I guess this is why testinfra encodes the string in general.
On Windows, a string means "the exact, escaped command, including arguments", and a list means "a command with args that needs to be escaped/joined". 

The fun comes on Windows as the encode() causes the following:
string "echo hello" passed to run_local()
string converted to b"echo hello"
b"echo hello" passed to subprocess.Popen
Popen checks the type, finds it's not a string, and runs list2cmdline()
list2cmdline expects a list, but gets a sequence of bytes, eventually blowing up at https://github.com/python/cpython/blob/v3.7.3/Lib/subprocess.py#L530 as iterating over a sequence of bytes leads to integers.

ianw: Thanks for the info, TIL about windows process creation functions and command-line encoding types, which I think everyone would be happier never having to think about ever :)

I tried to have a poke and see if anything obvious fell out, but I just ended up a bit more confused per https://github.com/philpep/testinfra/issues/411#issuecomment-521466951

Maybe this is right, but I suspect more that keeping this as "bytes" internally is wrong...

theothertom: Yeah, maybe "bytes" isn't correct internally. I guess some of this may come out with further testing - my team and I may well be doing a bit more work on Windows in the near future, so might end up doing some more stuff around getting the testinfra test suite working, which may reveal more things.

fatal-exception:
Can I ask what the current status is of this PR? What is the next action needed to merge? We're benefiting greatly from testinfra and are looking forward to this fix being merged in

keachi:
@theothertom @ianw What is the current status of this PR?

theothertom:
@keachi unfortunately, I've moved on to a new project and am no-longer in a position to test anything around this. Thinking back, I raised it as a patch posted in an issue comment was working for us, and it felt potentially useful to merge.
Given my lack of context, it's perhaps better for someone closer to https://github.com/pytest-dev/pytest-testinfra/issues/411 to raise a new PR that takes into account the last few years of changes.

PR #488, opened 2019-08-29
Add BlockDevice module to query block device info.

amarao:
I found I need some block device info for my tests, and testinfra got no such module, so I wrote it.

It is a wrapper over `blockdev --report` with some logic on top.

Gathered information:

* size
* r/w or r/o status
* sector_size
* block_size
* partition or not
* read ahead value
* basic sanity check (`is_block_device`)

### DIFF
--- a/testinfra/modules/blockdevice.py
+++ b/testinfra/modules/blockdevice.py
@@ -66,8 +66,9 @@ class BlockDevice(Module):
 
     @property
     def size(self):
-        """Return size if the device in bytes or None
-           if device is not a block device.
+        """Return size if the device in bytes.
+
+           Return None if device is not a block device.
 
         >>> host.block_device("/dev/sda1").size
         512110190592
--- a/testinfra/modules/blockdevice.py
+++ b/testinfra/modules/blockdevice.py
@@ -131,15 +131,15 @@ class LinuxBlockDevice(BlockDevice):
 
     @cached_property
     def _data(self):
-        HEADER = ['RO', 'RA', 'SSZ', 'BSZ', 'StartSec', 'Size', 'Device']
-        COMMAND = 'blockdev  --report %s'
-        blockdev = self.run(COMMAND % self.device)
+        header = ['RO', 'RA', 'SSZ', 'BSZ', 'StartSec', 'Size', 'Device']
+        command = 'blockdev  --report %s'
+        blockdev = self.run(command % self.device)
         if blockdev.rc != 0 or len(blockdev.stderr) > 0:
             raise RuntimeError("Failed to gather data: %s" % blockdev.stderr)
         output = blockdev.stdout.splitlines()
         if len(output) < 2:
             raise RuntimeError("No data from %s" % self.device)
-        if output[0].split() != HEADER:
+        if output[0].split() != header:
             raise RuntimeError('Unknown output of blockdev: %s' % output[0])
         fields = output[1].split()
         return {
--- a/testinfra/modules/blockdevice.py
+++ b/testinfra/modules/blockdevice.py
@@ -1,0 +1,223 @@
+# coding: utf-8
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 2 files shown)


amarao:
I've tried to write some tests, but Docker is not an option for block devices...

amarao:
I feel some of CI tests are failing due to external reasons...

testinfra/modules/blockdevice.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        if not self.is_block_device:
+            return None
+        return self._data['size']
+
+    @property
+    def sector_size(self):
+        """Return sector size for the device in bytes or None.
+
+        >>> host.block_device("/dev/sda1").sector_szie
```
philpep: typo sector_size

testinfra/modules/blockdevice.py
```diff
@@ -0,0 +1,220 @@
...
+        512
+
+        >>> print(host.block_device("/dev/console").sector_szie)
```
philpep: typo sector size

testinfra/modules/blockdevice.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        if not self.is_block_device:
+            return None
+        return self._data['sector_size']
+
+    @property
+    def block_size(self):
+        """Return block size for the device in bytes or None.
+
+        >>> host.block_device("/dev/sda").block_szie
```
philpep: same typo again ;)

testinfra/modules/blockdevice.py
```diff
@@ -0,0 +1,220 @@
...
+
+        >>> print(host.block_device("/dev/console").ra)
+        None
+        """
+        if not self.is_block_device:
+            return None
+        return self._data['read_ahead']
+
+    @property
+    def _data(self):
```
philpep: You can use the `@cached_property` decorator from `testinfra.utils` to avoid handling the `_data_cache` attribute.

testinfra/modules/blockdevice.py
```diff
@@ -0,0 +1,220 @@
...
+        if self._data_cache is None:
+            self._data_cache = self._probe_device(self.device)
+        return self._data_cache
+
+    @classmethod
+    def get_module_class(cls, host):
+        if host.system_info.type == 'linux':
+            return LinuxBlockDevice
+        # if host.system_info.type.endswith("bsd"):
+        #     return BSDlockDevice
```
philpep: NIT: drop this comment.

testinfra/modules/blockdevice.py
```diff
@@ -0,0 +1,220 @@
...
+    def __repr__(self):
+        return '<BlockDevice(path=%s)>' % self.device
+
+
+class LinuxBlockDevice(BlockDevice):
+
+    @classmethod
+    def _probe_device(cls, device):
+        HEADER = ['RO', 'RA', 'SSZ', 'BSZ', 'StartSec', 'Size', 'Device']
+        COMMAND = 'blockdev  --report %s'
```
philpep: Use lower-case variable names

testinfra/modules/blockdevice.py
```diff
@@ -0,0 +1,220 @@
...
+        run = cls(None).run
+        blockdev = run(COMMAND % device)
+        if blockdev.rc != 0:
```
philpep: I'm not sure hidding the error is a good idea here.
Maybe just use `check_output()` instead and let the error go to the user.
This could avoid falsely successful tests.

Same remark for `is_block_device`, I think the module should only be used with a block device, if not just raise an error.

On my machine (debian stretch) I have:

```
$ sudo blockdev --report /dev/console ; echo $?
RO    RA   SSZ   BSZ   StartSec            Size   Device
blockdev: ioctl error on /dev/console
0
```
(I'd expect rc != 0 but...)


So I'd suggest to detect this by splitting the second line and expect exactly 7 items.

testinfra/modules/blockdevice.py
```diff
@@ -0,0 +1,220 @@
...
+                'is_block_device': False,
+                'rw_mode': None,
+                'read_ahead': None,
+                'sector_size': None,
+                'block_size': None,
+                'start_sector': None,
+                'size': None
+            }
+        output = blockdev.stdout.splitlines()
+        assert output[0].split() == HEADER, 'Unknown output of blkid'
```
philpep: s/blkid/blockdev I guess

testinfra/modules/blockdevice.py
amarao: Oh, that was my big fault. I tested that --report does not give a proper exit code, but forgot about this. I will rework this piece completely.

Big thanks for finding this out early.

amarao:
Thank you a lot for feedback, I've reworked the code.

amarao:
Hello, @philpep, could you look into updated code, please?

philpep:
Merged, thanks @amarao !

Issue #458, opened 2019-06-26
addr.port().is_reachable do not work in CentOS 7


RainbowHerbicides:
If package `nmap-ncat` from default repo `centos-base` is present command `nc -w 1 -z` will not work due to absence of `-z` flag in version 6.40 and method `is_reachable` for port will always return `False`

Possible solution: check that `nc` version (`nc --version`) is not older than 7.25 (when -z option has been added)
https://github.com/nmap/nmap/pull/444
https://nmap.org/changelog  - mark [Ncat][GH#444]

philpep:
Maybe we should drop the `-z` here ?

julcollas:
By replacing all the time `nc -w 1 -z` by `echo /dev/zero | nc -w 1`should work in all OS

RainbowHerbicides:
I think that this issue can be fixed due to the simple fact that CentOS 7 is technically no more

PR #490, opened 2019-08-30
correct is_reachable to work with netcat and nmwap-ncat

julcollas:
fix issue #458

### DIFF
--- a/testinfra/modules/addr.py
+++ b/testinfra/modules/addr.py
@@ -32,7 +32,7 @@ class _AddrPort(object):
                 self._addr.name, self._port).rc == 0
 
         return self._addr.run(
-            "nc -w 1 -z %s %s", self._addr.name, self._port).rc == 0
+            "echo /dev/zero | nc -w 1 %s %s", self._addr.name, self._port).rc == 0
 
 
 class Addr(Module):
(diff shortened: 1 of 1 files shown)


testinfra/modules/addr.py
```diff
@@ -32,7 +32,8 @@ def is_reachable(self):
                 self._addr.name, self._port).rc == 0
 
         return self._addr.run(
-            "nc -w 1 -z %s %s", self._addr.name, self._port).rc == 0
+            "echo /dev/zero | nc -w 1 %s %s",
```
philpep: Can you explain how this works ? Actually this seems to send a string "/dev/zero" to the socket which is not equivalent to the `-z` option.

```
$ nc -w 1 -z 127.0.0.1 22
(nothing is send to sshd)
$ echo /dev/zero | nc -w 1 127.0.0.1 22  
SSH-2.0-OpenSSH_7.9p1 Debian-10
Protocol mismatch.
```

julcollas: You right. I see 2 possible solutions.

1. Never use **nc** and always use **bash**
2. Check if package **netcat** is present whereas **nc** binary

Any preferences ?

julcollas: up @philpep  :)

PR #307, opened 2018-04-29
add environment and x509

julcollas:
Environment module
x509 module

PR #491, opened 2019-08-31
add environment module

julcollas:
split #307

### DIFF
--- a/testinfra/modules/environment.py
+++ b/testinfra/modules/environment.py
@@ -1,0 +1,36 @@
+# coding: utf-8
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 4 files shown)


testinfra/modules/environment.py
```diff
@@ -0,0 +1,39 @@
...
+    {
+        "EDITOR": "vim",
+        "SHELL": "/bin/bash",
+        [...]
+    }
+    """
+
+    def __call__(self):
+        ret_val = dict(
+            i.split('=', 1) for i in self.check_output('env -0').splitlines()
```
philpep: Test is failing because you're using `splitlines()` instead of `split('\x00')`.

Also `env -0` isn't posix compliant https://pubs.opengroup.org/onlinepubs/9699919799/
For instance it doesn't work for busybox in alpinelinux. But it doesn't seems to exist a posix compliant command to do this.

ianw:
recheck

testinfra/modules/environment.py
julcollas: I've uploaded a correction for split

philpep:
Merged, thanks!

PR #508, opened 2019-10-25
add check parameter to run

randombenj:
Provide similar functionality like the [`subprocess.run(...)`](https://docs.python.org/3/library/subprocess.html#subprocess.run) command with the `check` parameter to raise an exception when the command failed to execute. This is useful when a command is not the test itself but used to prepare a test and you don't want to use an assert statement for every command.
Also I think this is more convenient than the `host.run_expect([0], ...)` command.

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -477,6 +477,13 @@ def test_command_execution(host):
     assert host.run("true").succeeded
 
 
+def test_run_check(host):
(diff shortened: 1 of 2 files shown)


philpep:
Hi, what about using existing `check_output()` for this instead ?

randombenj:
Yes `check_output` also throws an error but you don't get the whole result object returned.

PR #510, opened 2019-10-30
EditorConfig support

edubxb:
Add an EditorConfig configuration file to easy configure some code style
parameters.

+info: https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,26 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+insert_final_newline = true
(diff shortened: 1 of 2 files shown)


philpep:
The build is failed because we have to either include or exclude this new file from the python source distribution (sdist).

This can be done by adding "prune .editorconfig" in MANIFEST.in

Otherwise, this LGTM

edubxb:
Ok, sorry for that. Fixing it...

philpep:
Merged, thanks!

PR #511, opened 2019-10-30
Fix #505

### LINKED ISSUE #505: process.filter is unreliable with long usernames
https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.process.Process
suggests filtering processes by user `host.process.filter(user="root", comm="zsh")`

This fails with long usernames, e.g. as root
```
useradd user-with-long-name
su user-with-long-name -c 'sleep 1000'&
```
In Python:
```
In [1]: import testinfra

In [2]: host = testinfra.get_host('local://')

In [3]: host.process.filter(user='user-with-long-name')
Out[3]: []

In [4]: host.process.filter(user='user-wi+')
Out[4]: [<process sh (pid=4311)>, <process sleep (pid=4312)>]
```

The underlying call is `ps -Aww -o %s`:
https://github.com/philpep/testinfra/blob/3.2.0/testinfra/modules/process.py#L128
which may truncate the `user` field, though this depends on the order the fields are specified:
```
# ps -Aww -o pid,user,comm
  PID USER     COMMAND
    1 root     bash
 7315 root     ipython
 8029 root     su
 8036 user-wi+ sh
 8038 user-wi+ sleep
 8667 root     sh
 8668 root     ps
```
```
# ps -Aww -o pid,comm,user
  PID COMMAND         USER
    1 bash            root
 7315 ipython         root
 8029 su              root
 8036 sh              user-with-long-name
 8038 sleep           user-with-long-name
 8811 sh              root
 8812 ps              root
```
The list of attributes is a set so order is undefined: https://github.com/philpep/testinfra/blob/3.2.0/testinfra/modules/process.py#L129

I haven't found an easy fix, there don't seem to be any `ps` flags to ensure the full field is always output.

edubxb:
The -o flag in the ps command allow to specify the column width, so setting a fixed, and relatively wide value, should fix the issue.

### DIFF
--- a/testinfra/modules/process.py
+++ b/testinfra/modules/process.py
@@ -132,7 +132,7 @@ class PosixProcess(Process):
         attributes -= set(["lstart", "args"])
         attributes = sorted(attributes)
         attributes.extend(["lstart", "args"])
-        arg = ",".join(attributes)
+        arg = ":50,".join(attributes)
 
         procs = []
         # skip first line (header)


philpep:
This looks an awesome fix ! Thanks !

I just wonder if this is a portable solution on all "ps" implementations. Can you test this is accepted by "ps" on centos 6/7, debian stretch/buster, ubuntu xenial/bionic and maybe some BSD ? I'll test this ASAP.

edubxb:
Is tested in Debian Testing/Unstable/Stable that share the same ps version `3.3.15`, but I will try to test it on Debian Stretch and in CentOS 6/7 too. For BSD, I can't promise anything... :sweat_smile:

philpep:
This looks good on old distro too.

Merged, thanks!

PR #512, opened 2019-10-31
New limits module

edubxb:
Module to check [user session limits](http://man7.org/linux/man-pages/man5/limits.conf.5.html) using the [Bash ulimit builtin](https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html).

I not able to pass the test suite locally, some Docker builds fails on my laptop :sweat:.

### DIFF
--- a/doc/source/modules.rst
+++ b/doc/source/modules.rst
@@ -51,6 +51,10 @@ host
 
        :class:`testinfra.modules.iptables.Iptables` class
 
+    .. attribute:: limits
+
+       :class:`testinfra.modules.limits.Limits` class
+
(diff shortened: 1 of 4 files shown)


philpep:
Hi, thanks for this.

IIUC limits are per process and this module is exposing "default" limits applied to the current user.
I wonder if this could be integrated in the existing "process" module ? Process limits are exposed by linux in /proc/<pid>/limits.
And maybe this could also be exposed to the existing "user" module by running "cat /proc/self/limits" as selected user.

edubxb:
Mmm, sounds good, let me dig about it a little more.

PR #513, opened 2019-11-06
Implement File.list_dir to list items in a directory

farzadghanei:
So that users can assert on items in a directory. Just like the [File.content](https://github.com/philpep/testinfra/blob/02a9338b1e8eef7da8737099ee1445341e454fe1/testinfra/modules/file.py#L141) which is useful to assert if a file content contains a certain value, but for directory items.

### DIFF
--- a/testinfra/modules/file.py
+++ b/testinfra/modules/file.py
@@ -176,7 +176,7 @@ class File(Module):
         >>> host.file("/tmp").listdir
         ['foo_file', 'bar_dir']
         """
-        out = self.run_test("ls -- %s", self.path)
+        out = self.run_test("ls -1 -q -- %s", self.path)
         if out.rc != 0:
             raise RuntimeError("Unexpected output %s" % (out,))
         return out.stdout.splitlines()
--- a/testinfra/modules/file.py
+++ b/testinfra/modules/file.py
@@ -170,10 +170,10 @@ class File(Module):
         raise NotImplementedError
 
     @property
-    def list_dir(self):
+    def listdir(self):
         """Return list of items under the directory
 
-        >>> host.file("/tmp").list_dir
+        >>> host.file("/tmp").listdir
         ['foo_file', 'bar_dir']
         """
         out = self.run_test("ls -- %s", self.path)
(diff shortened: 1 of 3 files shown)


testinfra/modules/file.py
```diff
@@ -169,6 +169,18 @@ def size(self):
         """Return size of file in bytes"""
         raise NotImplementedError
 
+    @property
+    def list_dir(self):
```
philpep: Maybe `listdir()` is more straightforward for python people (like `os.listdir()`)

testinfra/modules/file.py
```diff
@@ -169,6 +169,18 @@ def size(self):
...
+        """Return list of items under the directory
+
+        >>> host.file("/tmp").list_dir
+        ['foo_file', 'bar_dir']
+        """
+        out = self.run_test("ls -- %s", self.path)
+        if out.rc != 0:
+            raise RuntimeError("Unexpected output %s" % (out,))
+        return out.stdout.splitlines()
```
philpep: I think this will not work for files containing spaces (or even \r or \n) in their name. Maybe `ls -1` (and additional option) or `find` command has a way to parse them easily, maybe using \0 (null) separator ?

testinfra/modules/file.py
farzadghanei: I agree resemblance to Python standard library could be a plus. I followed the convention from current module ([is_file](https://github.com/philpep/testinfra/blob/392a3b8f5e958a152b8b34f1fb7a907b5c67d6cc/testinfra/modules/file.py#L42), [is_directory](https://github.com/philpep/testinfra/blob/392a3b8f5e958a152b8b34f1fb7a907b5c67d6cc/testinfra/modules/file.py#L46), etc.), all functions are snake_cased and none named as attachedwords. Do you still prefer `list_dir` to be renamed to `listdir`?

philpep: I'd prefer listdir (not a strong opinion through)

testinfra/modules/file.py
farzadghanei: very good point. I used `ls -q` to handle new lines (7f80bb13c24a7d1dc092a4b690f4c528e04ffd85 commit message tells more)

philpep:
Merged, thanks @farzadghanei ! (and sorry for delay...)

farzadghanei:
@philpep Thanks for developing and maintaining this great tool. :rocket:

PR #514, opened 2019-11-17
ansible + ssh backend: support for passwords via sshpass

sumkincpp:
If ssh identity file is not provided via `--ssh-identity-file` flag, testinfra will try 
to use `ansible_ssh_private_key_file`, `ansible_private_key_file` and, finally, 
`ansible_user` with `ansible_ssh_pass` variables, both should be specified.


philpep:
Merged, thanks for this !

sumkincpp:
@philpep Cool, thanks!

PR #525, opened 2019-12-08
Remove --tree None option from ansible runner

francoisfreitag:
`None` has been the default value for tree since the option was added,
https://github.com/ansible/ansible/commit/ab17f6f44e8fc3c15549b6ff4b5c752ef1a79bd9


francoisfreitag:
Nevermind, tree is used by https://github.com/philpep/testinfra/blob/1ea349e94795dac36d82d56d229b9876044ace73/testinfra/utils/ansible_runner.py#L217

PR #526, opened 2019-12-08
Allow specifying arguments to the Ansible CLI

francoisfreitag:
Instead of adding a keyword argument for each supported option, allow users to pass arguments directly to the ansible command line.

Support for passing [extra vars](https://docs.ansible.com/ansible/latest/cli/ansible.html#cmdoption-ansible-e) has been proposed in https://github.com/philpep/testinfra/pull/462.

This approach offers support for extra vars, as well as become user, verbosity, running in diff mode, etc.

### DIFF
--- a/test/test_backends.py
+++ b/test/test_backends.py
@@ -355,6 +355,53 @@ def test_ansible_config():
                 del os.environ['ANSIBLE_CONFIG']
 
 
+@pytest.mark.parametrize(
+    'options,expected_cli,expected_args',
+    [
+        ({}, "--check", []),
+        ({"become": True}, "--become --check", []),
+        ({"check": False}, "", []),
+        ({"diff": True, "check": False}, "--diff", []),
+        ({"one_line": True, "check": False}, "--one-line", []),
+        (
+            {"become_method": "sudo", "check": False},
+            "--become-method %s",
+            ["sudo"],
+        ),
+        (
+            {"become_user": "root", "check": False},
+            "--become-user %s",
+            ["root"],
+        ),
+        ({"user": "root", "check": False}, "--user %s", ["root"]),
+        (
+            {
+                "extra_vars": {"target": "production", "foo": 42},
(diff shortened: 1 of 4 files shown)


francoisfreitag:
I wonder if `become` has some value over passing `--become` to `cli_options`? If not, perhaps it could be deprecated?

philpep:
Hi, thanks for this, this looks definitively useful !

I wonder if we could write some wrapper to transform a kwargs dict to command line option. This would allow to pass extra options without needing to handle shell quoting issues.

```
to_options({'check': True}) -> --check
to_options({'become_user': 'nobody'}) -> --become-user nobody
to_options({'extra_vars': {'foo': 'bar'}) -> --extra-vars '{"foo": "bar"}'
```

What do you think ?

hglib has such code that we could just copy into testinfra;utils:cmdbuilder: https://www.mercurial-scm.org/repo/python-hglib/file/tip/hglib/util.py#l79

francoisfreitag:
Extra vars requires some special logic (to serialize the provided dict value as JSON), which the `cmdbuilder` from hglib does not handle.

I tried to pilot ansible using its API instead of the CLI, looking at https://github.com/ansible/ansible/blob/cd95843ea5f1bff40225b0430cfbb379cbe53661/lib/ansible/cli/adhoc.py for reference. IIUC, the ansible runner only accepts options via CLI, making it hard to instrument ansible directly from a controlling python process.

I like the idea of validating and formatting options before passing them to ansible, so that users can pass their arguments in a richer way than a plain string. It gives the library some liberty to mangle with the input (e.g. serializing extra vars value to JSON).

Listing available options in the ansible runner module, and validating inputs against that list seems a good path forward. I’ll give that a shot.

francoisfreitag:
PR updated to accept Ansible options as keyword arguments.

philpep:
Merged, thanks @francoisfreitag !

PR #527, opened 2019-12-16
Add preliminary support for podman.

larsks:
Podman is another container management tool, similar in function to
Docker (but different in operation).  This adds a podman module, a
podman backend, and updates the docs accordingly.


philpep:
Merged ! Thanks for this !

PR #532, opened 2020-01-27
Add support for Windows files and services to TestInfra

fatal-exception:
This PR, which also contains changes in https://github.com/philpep/testinfra/pull/480 which we needed to get testinfra to work on windows, adds support for Windows files and services, allowing tests to be written against these objects.

I was having trouble running the tests (they seem to require a French locale), so I'm hoping Travis will check this PR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ flake8.report
 junit*.xml
 doc/build
 .cache
+.idea
--- a/testinfra/modules/file.py
+++ b/testinfra/modules/file.py
@@ -189,6 +189,8 @@ class File(Module):
             return BSDFile
         if host.system_info.type == "darwin":
             return DarwinFile
+        if host.system_info.type == "windows":
+            return WindowsFile
         raise NotImplementedError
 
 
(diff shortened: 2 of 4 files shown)


testinfra/modules/file.py
```diff
@@ -303,3 +307,167 @@ class NetBSDFile(BSDFile):
...
+        >>> host.file("/etc/password").mode == stat.S_IRUSR | stat.S_IWUSR
+        True
+
+        .. _oct(x): https://docs.python.org/3.5/library/functions.html#oct
+        .. _stat: https://docs.python.org/2/library/stat.html
+        """
+        raise NotImplementedError
+
+    def contains(self, pattern):
+        return self.run_test("grep -qs -- %s %s", pattern, self.path).rc == 0
```
philpep: I guess this should be NotImplementedError

testinfra/modules/file.py
```diff
@@ -303,3 +307,167 @@ class NetBSDFile(BSDFile):
...
+    @property
+    def md5sum(self):
+        raise NotImplementedError
+
+    @property
+    def sha256sum(self):
+        raise NotImplementedError
+
+    def _get_content(self, decode):
+        out = self.run_test("cat -- %s", self.path)
```
philpep: does windows has cat command ?

testinfra/modules/file.py
```diff
@@ -303,3 +307,167 @@ class NetBSDFile(BSDFile):
...
+
+    def __eq__(self, other):
+        if isinstance(other, File):
+            return self.path == other.path
+        if isinstance(other, six.string_types):
+            return self.path == other
+        return False
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
```
philpep: Please do not redefine function already defined in parent class

testinfra/modules/file.py
KosherSalt: Only in PowerShell. It's an alias for Get-Content
The equivalent in cmd would be type
https://ss64.com/nt/type.html

starlord-daniel: Is this resolved?

.gitignore
```diff
@@ -18,3 +18,4 @@ flake8.report
 junit*.xml
 doc/build
 .cache
+.idea
```
atombrella, line 21: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

tdas3001: Thanks for opening this up once again! Having windows support for testinfra is a really nice feature. Is it planned to release the feature soon? Also, is there any workaround till then? Any comment or suggestion is appreciated!

keachi:
@fatal-exception @philpep What is the status with this PR?

testinfra/modules/file.py
```diff
@@ -12,6 +12,8 @@
 # limitations under the License.
 
 import datetime
+import six
```
atombrella, line 15: Only Python 3 support. Drop this!

tdas3001:
@fatal-exception @philpep any progress on the PR?

philpep:
@tdas3001 I've no time and no knowledge to review or maintain this. So unless there's a new maintainer able to handle it, windows support will remain minimal.

PR #543, opened 2020-03-05
Make File.listdir a method

dlax:
This documented as a method in 4.1 changelog and it seems more intuitive
this way as it clearly indicates that a computation is performed.


philpep:
Merged, thanks!

PR #545, opened 2020-04-09
Respect remote_user and remote_port in ansible.cfg

iafm:
Currently `get_ansible_host()` ignores remote_user and remote_port variables from ansible.cfg. This means that if you're using non-default ssh port or if your local username is different from the remote username, you have to set ansible_port or ansible_user variables in your ansible inventory, which is often inconvenient, especially when you're using shared inventory file.


philpep:
This LGTM, merged, thanks!

PR #550, opened 2020-05-07
Add architecture to sysinfo

ianw:
This allows you to switch on the host architecture easily.  "uname -m"
seems to be globally acceptable on unix-y hosts, Windows has
%PROCESSOR_ARCHITECTURE%.


philpep:
Merged, thanks!

PR #557, opened 2020-07-02
Add docstrings to User and Group .exists properties.

disser:


philpep:
Merged, thanks @disser !

PR #561, opened 2020-07-22
service-module: use /run/systemd/system/ to detect systemd (fixes #546)

### LINKED ISSUE #546: Service module not working on openSUSE Leap
Hi!

First of all, thank you for your work on this project! I'm using it with ansible and molecule to test my custom roles before deploying them to my production machines and I found testinfra to be the ideal tool to specify my test criteria. 

Unfortunately, the service module does not work on recent SUSE/openSUSE installations. SLES switched from SysVinit to systemd in version 12 and openSUSE followed in 12.1. However, testinfra reckons it is using OpenRC and therefore fails with the following message:
`AssertionError: Unexpected exit code 1 for CommandResult(command=b"find -L /etc/rc?.d/ -name 'S??postfix'", exit_status=1, stdout=None, stderr=b"find: '/etc/rc?.d/': No such file or directory\n")`

I looked into the code and found these lines:
https://github.com/philpep/testinfra/blob/0e337bb02bc4df10799f2dc7af106f4d0a64dc10/testinfra/modules/service.py#L50-L54

The check fails because there is no `/sbin/init` on an openSUSE system. I suggest to follow the ~~inofficial~~ official documentation for the [sd_booted](https://www.freedesktop.org/software/systemd/man/sd_booted.html) function and to check, whether the directory `/run/systemd/system/` exists. That should be a more portable solution to this problem.

I can file a pull-request for this change, if it helps. 

Cheers!

ckanibal:
I couldn't quite get the unit tests working. I'll leave that to Travis. The change is designed to not break anything :-)

### DIFF
--- a/testinfra/modules/service.py
+++ b/testinfra/modules/service.py
@@ -45,7 +45,7 @@ class Service(Module):
     @classmethod
     def get_module_class(cls, host):
         if host.system_info.type == "linux":
-            if (
+            if host.file("/run/systemd/system/").is_directory or (
                 host.exists("systemctl")
                 and "systemd" in host.file("/sbin/init").linked_to
             ):


philpep:
Merged, thanks!

PR #398, opened 2019-01-22
Decode issues when using salt host

jynolen:
Command run
`/opt/pytest/bin/py.test -v --host="salt://master-*" /opt/terraform/test/salt/test_infra/test.py`

Content of /opt/terraform/test/salt/test_infra/test.py
```
def test_salt_master(host):
    assert host.socket("tcp://:::4506").is_listening

def test_salt_clients_minions(host):
    assert len(json.loads(host.run("salt-run --out-indent=-1 --out=json manage.up minion-*").stdout)) == count_minion
```

Expected : Test Passed

Result :

```def test_salt_master(host):
>       assert host.socket("tcp://:::4506").is_listening

/opt/terraform/test/salt/test_infra/test_splunk_network.py:7: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/opt/pytest/lib/python3.4/site-packages/testinfra/modules/socket.py:112: in is_listening
    sockets = list(self._iter_sockets(True))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <socket tcp://:::4506>, listening = True

    def _iter_sockets(self, listening):
        cmd = '%s --numeric'
        if listening:
            cmd += ' --listening'
        else:
            cmd += ' --all'
        if self.protocol == 'tcp':
            cmd += ' --tcp'
        elif self.protocol == 'udp':
            cmd += ' --udp'
        elif self.protocol == 'unix':
            cmd += ' --unix'
    
        for line in self.run(cmd, self._command).stdout_bytes.splitlines()[1:]:
[...]

PR #566, opened 2020-08-10
Fix #398 `salt` backend returning non-encoded strings

salty-vagrant:
`salt` backend failed because local salt client returns
`stdout` and `stderr` as `str` objects not encoded `bytes`.

Taking a lead from the `ansible` implementation the `stdout` and
`stderr` results are both encoded and returned as `kwargs` through
`base.result` method.


philpep:
Hi, thanks for this!

What version of salt are you using and using which python version (on salt side) ? I guess this does not occur always because we have a test which should fail, so maybe it's related by python2 salt ?

scornelissen85:
I can confirm that this has to be merged for Salt 3000 or newer to work.

scornelissen85:
This can be closed, as it is merged in #636.

Issue #567, opened 2020-08-12
Add is_masked property on Service (or directly in SystemdService)


dorianm:
Hi,

I need to test if a systemd service is *masked* (see : [`systemctl mask`](https://www.freedesktop.org/software/systemd/man/systemctl.html#mask%20UNIT…)). Currently, the `Service` class only have two properties `is_enabled` and `is_running`.  

When I checked the code, I saw the `SystemdService` class implements a `is_valid` property (which is not documented because it's not a property of the `Service` class).

*First question*: should we put the `is_valid` property in the `Service` class to complete the documentation ?
*Second question*: may I implement the `is_masked` property ? If yes, should I put the property only on the `SystemdService` our should I put it in the `Service` class (I think the second solution is better).

Thanks you in advance.

PR #568, opened 2020-09-02
Fix#567 : Implement is_masked in Service and add doc for is_masked and is_valid

dorianm:
Hi,

See #567 for the context of changes.

And see that comment from 2017 for the documentation part : https://github.com/philpep/testinfra/pull/265#issuecomment-346306163


philpep:
Merged, thanks for this !

I added a followup https://github.com/philpep/testinfra/commit/2035db0338664bec3bb5fc8e51e263697557a31e and tagged 5.3.0 which should be released to pypi in a few minutes.

PR #569, opened 2020-09-03
Fix is_masked method on systemd service 

dorianm:
Hi

I'm sorry but I introduced a bug with my PR #568. The `systemcl is-enabled` command doesn't return 0 when the service is not enabled. So my check `cmd.rc == 0` was not relevant in this case.

Sorry for this bug. I add on test to check if my code is valid. This time, it should be ok !

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -133,20 +133,13 @@ def test_ssh_service(host, docker_image):
         assert ssh.is_enabled
 
 
-@all_images
-def test_service_mask(host, docker_image):
-    if docker_image in ("centos_6", "centos_7",
-                        "alpine", "archlinux"):
-        name = "sshd"
-    else:
-        name = "ssh"
-
-    ssh = host.service(name)
-    if isinstance(ssh, SystemdService):
-        host.run("systemctl mask %s", name)
-        assert ssh.is_masked
-        host.run("systemctl unmask %s", name)
-        assert not ssh.is_masked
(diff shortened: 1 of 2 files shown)


test/test_modules.py
```diff
@@ -132,6 +133,22 @@ def test_ssh_service(host, docker_image):
...
+        name = "sshd"
+    else:
+        name = "ssh"
+
+    ssh = host.service(name)
+    if isinstance(ssh, SystemdService):
+        host.run("systemctl mask %s", name)
+        assert ssh.is_masked
+        host.run("systemctl unmask %s", name)
+        assert not ssh.is_masked
```
philpep, line 149:
```suggestion
def test_service_systemd_mask(host):
    ssh = host.service("ssh")
    assert not ssh.is_masked
    host.run("systemctl mask %s", name)
    assert ssh.is_masked
    host.run("systemctl unmask %s", name)
    assert not ssh.is_masked
```

test/test_modules.py
```diff
@@ -20,6 +20,7 @@
 from ipaddress import ip_address
 from ipaddress import IPv4Address
 from ipaddress import IPv6Address
+from testinfra.modules.service import SystemdService
```
philpep, line 23:
```suggestion
```

philpep:
Merged with some simplification of the test. Thanks! This will released as 5.3.1 asap

PR #571, opened 2020-10-15
utils/ansible_runner: Use ssh_args from ansible.cfg

BenoitKnecht:
If `ssh_args` is set in Ansible's configuration file, use those arguments when
connecting to hosts.

Signed-off-by: Benoît Knecht <bknecht@protonmail.ch>


philpep:
Merged, thanks!

PR #572, opened 2020-10-15
backend: Deduplicate hosts returned by get_backends()

BenoitKnecht:
If a host appears multiple times in host specs (which can typically happen if a
host is a member of multiple Ansible groups for instance), testinfra would run
tests on that host multiple times.

This commit turns the list of hosts that are being considered into a set, and
makes each one of them hashable in such a way that two hosts are considered
equal if they have the same name, connection parameters, etc.

As a result, if a host finds itself listed multiple times with the same backend
and connection parameters, testinfra will run the tests on that host only once.

Signed-off-by: Benoît Knecht <bknecht@protonmail.ch>

### DIFF
--- a/test/test_backends.py
+++ b/test/test_backends.py
@@ -519,3 +519,16 @@ def test_ssh_hostspec(hostspec, expected):
     cmd, cmd_args = backend._build_ssh_command('true')
     command = backend.quote(' '.join(cmd), *cmd_args)
     assert command == expected
+
+
+def test_get_hosts():
+    # Hosts returned by get_host must be deduplicated (by name & kwargs) and in
+    # same order as asked
+    hosts = testinfra.backend.get_backends([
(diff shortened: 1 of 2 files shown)


philpep:
Hi, thanks for this and sorry for late reply.

I made some modifications:

* get_backends() should return hosts in the same order as requested (hence use a dict which is sorted by insertion with py36+ whereas sets are not).
* Use a cache key instead of implementing __hash__ in each backend. I found implementing __hash__ quite complex and may lead to future errors when adding new attributes. Note my solution has a drawback over yours of duplicating hosts in case we supply a parameter with its default value. But I think this is an edge case.
* Add a test

I'll merge as soon as the CI is ok.

Thanks!

BenoitKnecht:
No worries, thanks for the review and improvements! I'm glad to see this feature merged. :slightly_smiling_face:

Issue #570, opened 2020-09-11
is_installed fails in a RpmPackage that has dpkg installed


guimaluf:
I'm testing a CentOS system and after installing `dpkg` package in that system, testinfra tests fail for all packages.
[Looking at the code](https://github.com/philpep/testinfra/blob/master/testinfra/modules/package.py#L64-L79)
```python    
    @classmethod
    def get_module_class(cls, host):
        if host.system_info.type == 'windows':
            return ChocolateyPackage
        if host.system_info.type == "freebsd":
            return FreeBSDPackage
        if host.system_info.type in ("openbsd", "netbsd"):
            return OpenBSDPackage
        if host.exists("dpkg-query"):
            return DebianPackage
        if host.exists("rpm"):
            return RpmPackage
        if host.exists("apk"):
            return AlpinePackage
        if host.system_info.distribution == "arch":
            return ArchPackage
        raise NotImplementedError
```

I can see that instead of deciding the `Package` class on `system_info.distribution` the decision to use dpkg is based solely in its presence. 

## How to reproduce
In a CentOS host, install `dpkg` package and check for any package `host.package("kernel").is_installed`
```
(Pdb) host.package("kernel").version
*** AssertionError: Unexpected exit code 127 for CommandResult(command=b"dpkg-query -f '${Status} ${Version}' -W munge", exit_status=127, stdout=None, stderr=b"Warning: Permanently added '10.43.102.95' (ED25519) to the list of known hosts.\r\nbash: dpkg-query: command not found\n")
assert 127 == 0
```

guimaluf:
This is how I've fixed. Unfortunately, right now I don't have the bandwidth to submit a merge request. :/
```python
...
        if host.system_info.distribution in ("debian","ubuntu"):
            return DebianPackage
        if host.system_info.distribution in ("centos"):
            return RpmPackage
...
```

PR #575, opened 2020-10-26
Fix module class for RPM and Debian distributions #570

guimaluf:
A CentOS system with dpkg package installed fails for all packages
is_installed checks. This happens because the decision to use the module class
is based on the presence of `dpkg-query` binary.

This commit changes the check to look for `system_info.distribution`, and
makes CentOS 6 distribution name lower case to be consistent.

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -85,7 +85,6 @@ def test_non_default_package_tool(host):
     # Make non default pkg tool binary present
     host.run("install -m a+rx /bin/true /usr/bin/dpkg-query")
     # drop the cache
-    del host.package
     assert host.package("openssh").is_installed
 
 
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -31,6 +31,13 @@ all_images = pytest.mark.testinfra_hosts(*[
     )
 ])
 
+centos_images = pytest.mark.testinfra_hosts(*[
+    "docker://{}".format(image)
+    for image in (
+        "centos_6", "centos_7",
+    )
+])
+
 
 @all_images
 def test_package(host, docker_image):
@@ -73,6 +80,16 @@ def test_held_package(host):
 
 
 @pytest.mark.destructive
+@centos_images
+def test_non_default_package_tool(host):
+    # Make non default pkg tool binary present
+    host.run("install -m a+rx /bin/true /usr/bin/dpkg-query")
+    # drop the cache
+    del host.package
+    assert host.package("openssh").is_installed
+
+
+@pytest.mark.destructive
 def test_uninstalled_package_version(host):
     with pytest.raises(AssertionError) as excinfo:
         host.package('zsh').version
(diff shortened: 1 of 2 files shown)


philpep:
Hi, thanks for this !

However I think this have to be more complex because there are toons of centos/debian variants (like fedora, mint etc) and they have all a different "ID" in /etc/os-release.

When installing debian tools on centos, do you also have the apt binaries (e.g. apt-cache) ? If not this might be a good criteria.

Or we could test both existence of dpkg-query and rpm and handle the case where both are present using system_info as criteria like in your patch, handling known distributions for now.

guimaluf:
Hi philpep, 
thanks for your attention. I agree with you.
What if we keep the fallback condition  checking dpkg-query and rpm binaries, not the neatest code but might do the work.

test/test_modules.py
```diff
@@ -72,6 +78,14 @@ def test_held_package(host):
     assert python.version.startswith("2.7.")
 
 
+@pytest.mark.destructive
+@centos_images
+def test_non_default_package_tool(host):
+    # Make non default pkg tool binary present
+    host.run("install -m a+rx /bin/true /usr/bin/dpkg-query")
```
philpep, line 85:
```suggestion
    host.run("install -m a+rx /bin/true /usr/bin/dpkg-query")
    # drop the cache
    del host.package
```

testinfra/modules/systeminfo.py
```diff
@@ -88,7 +88,7 @@ def _get_linux_sysinfo(self):
         if redhat_release.rc == 0:
             match = re.match(
                 r"^(.+) release ([^ ]+) .*$",
-                redhat_release.stdout.strip())
+                redhat_release.stdout.strip().lower())
```
philpep, line 91:
```suggestion
                redhat_release.stdout.strip()
```

testinfra/modules/package.py
```diff
@@ -68,14 +68,19 @@ def get_module_class(cls, host):
             return FreeBSDPackage
         if host.system_info.type in ("openbsd", "netbsd"):
             return OpenBSDPackage
-        if host.exists("dpkg-query"):
+        if host.system_info.distribution in ("debian", "ubuntu"):
             return DebianPackage
-        if host.exists("rpm"):
+        if host.system_info.distribution == "centos":
```
philpep, line 73:
```suggestion
        if host.system_info.distribution and host.system_info.distribution.lower() == "centos":
```

test/test_modules.py
```diff
@@ -93,7 +110,7 @@ def test_systeminfo(host, docker_image):
     release, distribution, codename, arch = {
         "alpine": (r"^3\.11\.", "alpine", None, 'x86_64'),
         "archlinux": ("rolling", "arch", None, 'x86_64'),
-        "centos_6": (r"^6", "CentOS", None, 'x86_64'),
+        "centos_6": (r"^6", "centos", None, 'x86_64'),
```
philpep, line 113:
```suggestion
        "centos_6": (r"^6", "CentOS", None, 'x86_64'),
```

test/test_modules.py
```diff
@@ -72,6 +79,16 @@ def test_held_package(host):
     assert python.version.startswith("2.7.")
 
 
+@pytest.mark.destructive
+@centos_images
+def test_non_default_package_tool(host):
+    # Make non default pkg tool binary present
+    host.run("install -m a+rx /bin/true /usr/bin/dpkg-query")
+    # drop the cache
+    del host.package
```
philpep, line 88:
```suggestion
```

test/test_modules.py
```diff
@@ -72,6 +79,15 @@ def test_held_package(host):
     assert python.version.startswith("2.7.")
 
 
+@pytest.mark.destructive
+@centos_images
+def test_non_default_package_tool(host):
+    # Make non default pkg tool binary present
+    host.run("install -m a+rx /bin/true /usr/bin/dpkg-query")
+    # drop the cache
```
philpep, line 87:
```suggestion
```

philpep:
Merged with minor changes, thanks!

PR #579, opened 2020-11-02
Remove "encoding" parameter when binary mode is used

pilou-:
Fix this exception:

    python3 -c "import sys; from testinfra.plugin import SpooledTemporaryFile;\
                SpooledTemporaryFile(encoding=sys.stdout.encoding).fileno()"
      File "<string>", line 1, in <module>
      File "/usr/lib/python3.8/tempfile.py", line 839, in fileno
        self.rollover()
      File "/usr/lib/python3.8/tempfile.py", line 793, in rollover
        newfile = self._file = TemporaryFile(**self._TemporaryFileArgs)
      File "/usr/lib/python3.8/tempfile.py", line 744, in TemporaryFile
        return _io.open(fd, mode, buffering=buffering,
    ValueError: binary mode doesn't take an encoding argument

### DIFF
--- a/testinfra/plugin.py
+++ b/testinfra/plugin.py
@@ -167,7 +167,10 @@ class NagiosReporter:
 class SpooledTemporaryFile(tempfile.SpooledTemporaryFile):
 
     def __init__(self, *args, **kwargs):
-        self._out_encoding = kwargs['encoding']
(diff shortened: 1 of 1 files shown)


philpep:
Merged, thanks!

PR #584, opened 2020-11-06
Add support for password based authorization with paramiko

MartinWelcker:
Authorization with username and password is now possible.
The password attribute will now get passed to Paramiko. 
As I noticed it just seems to be a missing configuration value.
Users can now authenticate using connection strings like the following one:
```python
testinfra.get_host(F"paramiko://{username}:{password}@{address}:{port}")
```


philpep:
Merged, thanks!

PR #595, opened 2020-12-23
Fix #594: mode documentation

### LINKED ISSUE #594: Example for mode is wrong
It seems that the example for https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.file.File.mode is wrong.

Current:
```
>>> host.file("/etc/passwd").mode
384  # 0o600 (octal)
>>> host.file("/etc/password").mode == 0o600
True
>>> oct(host.file("/etc/password").mode) == '0600'
True
```

Should be:
```
>>> host.file("/etc/passwd").mode
384  # 0o600 (octal)
>>> host.file("/etc/password").mode == '0600'
True
>>> oct(host.file("/etc/password").mode) == 0o600
True
```

romainx:
Hello,

Fix #594 a small issue in the `File.mode` documentation, see my [comment][1] in the issue for more detail.

Best.

[1]: https://github.com/pytest-dev/pytest-testinfra/issues/594#issuecomment-750370503

### DIFF
--- a/testinfra/modules/file.py
+++ b/testinfra/modules/file.py
@@ -97,7 +97,7 @@ class File(Module):
         384  # 0o600 (octal)
         >>> host.file("/etc/password").mode == 0o600
         True
-        >>> oct(host.file("/etc/password").mode) == '0600'
+        >>> oct(host.file("/etc/password").mode) == '0o600'
         True
 
         Note: Python 3 oct(x)_ function will produce ``'0o600'``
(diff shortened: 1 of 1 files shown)


philpep:
Hi, thanks for this! Actually the difference comes from python2 vs python3.

So I added a fix:

* drop comment about python2 (now we only use python3)
* update urls to python2
* use a working example using /etc/shadow

romainx:
@philpep thanks for that. 
I had just done a very basic fix without asking me a lot of questions. I will do better next time 😄

philpep:
Merged, thanks!

PR #596, opened 2020-12-26
First implementation of network namespaces in addr module

### LINKED ISSUE #559: Linux network namespace support for addr module
Hello.

Problem:
addr module actions don't support Linux network namespaces

It would be nice to add Linux network namespaces support by running commands (ping, nc) with `ip net exec ns_name command` prefix

romainx:
Hello,

Here is a try to resolve #559 by supporting network namespaces in the `addr` module.

It's not easy to test since it requires privileged mode in docker and it's not possible (as far as I know) to setup a namespace in a Docker image at build time. So I have only made some minimal test.

I'm not sure of the way to handle the case when `ip` command is not available. The current implementation throws immediately a `NotImplementedError` in this case.

Note: In some cases I've used the `format` method to format string since I had trouble using the classic `%` way. I don't know if it's a problem or if we need to stick to one convention.

Best.

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -588,3 +588,15 @@ def test_addr(host):
 
     for ip in google_addr.ip_addresses:
         assert isinstance(ip_address(ip), (IPv4Address, IPv6Address))
+
+
+@pytest.mark.testinfra_hosts("ansible://debian_buster")
+def test_addr_namespace(host):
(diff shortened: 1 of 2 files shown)


philpep:
Hi @romainx , sorry to not have a look on this before. Code looks fine to me so it's merged, thanks!

PR #605, opened 2021-03-06
pip check support in PipPackage module

romainx:
Hello,

This PR adds [`pip check`](https://pip.pypa.io/en/stable/reference/pip_check/) support in `PipPackage` module.

> Verify installed packages have compatible dependencies.

Best

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -510,6 +510,7 @@ def test_pip_package(host):
         pip_path='/v/bin/pip')['pytest']
     assert outdated['current'] == pytest['version']
     assert int(outdated['latest'].split('.')[0]) > 2
+    assert host.pip_package.check().succeeded
 
 
 def test_environment_home(host):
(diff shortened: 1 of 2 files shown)


romainx:
The error is `test_ssh_service` on py36 / alpine, seems not related to this PR.

philpep:
Merged, thanks!

romainx:
Thank you @philpep I would like to ask you a question. 
Imagine we would like to implement something more related to the `pip` package manager than the packages themselves (for example related to `pip` configuration). Do you think it should be implemented in the `PipPackage` `class` or do you think it should be implemented in another `class` (for example called `Pip`) and dedicated to the package manager itself?

romainx:
@philpep thanks for the quick answer. Your suggestion makes sense. I share the same opinion and I think performing this change would make the things clearer. If you're OK, I will have a deepest look at it and propose something that remains compliant with the current implementation. I will manage it in a new issue + PR. Thanks a lot for this exchange, really appreciate.

PR #606, opened 2021-03-14
pip refactoring: implementation of installed and version

romainx:
Hello,

As discussed in #605 rename `pip_package` to `pip` and make it works like `package` module e.g:

```python
# instance
django = pip('Django')
assert django.is_installed
assert django.version.startswith('3.1')
# class method
assert pip.check()
```

First implementation of `is_installed` and `version` as properties.

So `Pip` is now a `Module` instead of `InstanceModule` and API not related to a package are implemented as a `classmethod`.

`pip_package` is deprecated with a proper warning (in the code and in the doc) and still works as before.

_shell escaped_ every `pip` calls as suggested by @philpep.

Best

### DIFF
--- a/testinfra/modules/pip.py
+++ b/testinfra/modules/pip.py
@@ -39,7 +39,7 @@ class Pip(Module):
         >>> host.package("pip").is_installed
         True
         """
-        return self.run_test("{} show %s".format(self.pip_path), self.name).rc == 0
+        return self.run_test("%s show %s", self.pip_path, self.name).rc == 0
 
     @property
     def version(self):
@@ -49,7 +49,8 @@ class Pip(Module):
         '18.1'
         """
         return self.check_output(
-            "{} show %s | grep Version: | cut -d' ' -f2".format(self.pip_path),
+            "%s show %s | grep Version: | cut -d' ' -f2",
+            self.pip_path,
             self.name,
         )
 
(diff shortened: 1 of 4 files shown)


testinfra/modules/pip.py
```diff
@@ -23,10 +24,37 @@ def _re_match(line, regexp):
...
+        super().__init__()
+
+    @property
+    def is_installed(self):
+        """Test if the package is installed
+
+        >>> host.package("pip").is_installed
+        True
+        """
+        return self.run_test("{} show %s".format(self.pip_path), self.name).rc == 0
```
philpep, line 42: I think `self.pip_path` should be eventually shell escaped (in case of spaces or shell restricted characters). This can be done in a followup PR or commit.

```
return self.run_test("%s show %s", self.pip_path, self.name).rc == 0
```

testinfra/modules/pip.py
```diff
@@ -100,11 +130,40 @@ def get_outdated_packages(self, pip_path="pip"):
...
+    def _deprecated():
+        warnings.warn(
+            "Calling host.pip_package is deprecated, call host.pip instead",
+            DeprecationWarning,
+        )
+
+    @classmethod
+    def check(cls, pip_path="pip"):
+        PipPackage._deprecated()
+        return super(PipPackage, cls).check(pip_path=pip_path)
```
philpep, line 159: We are python3 only so you can write `` return super().check(pip_path=pip_path)`` here and in following code.

philpep (changes requested):
Awesome, thanks for this !

I left some comments.

testinfra/modules/pip.py
romainx: Right, thanks, it's done.

testinfra/modules/pip.py
romainx: Done for all calls.

philpep:
Merged, thanks!

PR #607, opened 2021-03-18
Add a supervisorctl path option and ability to pass a configuration file path 

### LINKED ISSUE #536: Add a supervisorctl path option and ability to pass a configuration file path
Hello!,

Just started working with this package and have to say is great!.

Was playing with the [Supervisor](https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.supervisor.Supervisor) class and would love to have an option specify the path where `supervisorctl` resides as well to be able to define where the configuration may reside.

There are some scenarios where the installation and configuration might not be in the system standard path.

Like:

```shell
/opt/supervisor/bin/supervisorctl -c /another/path/supervisord.conf status <some service>
```

romainx:
Hello

Closes #536.

```python
gunicorn = host.supervisor("gunicorn", "/usr/bin/supervisorctl", "/etc/supervisor/supervisord.conf")
assert gunicorn.status == "RUNNING"
```

Best

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -423,7 +423,14 @@ def test_ansible_module_options(host):
 
 
 @pytest.mark.destructive
-def test_supervisor(host):
+@pytest.mark.parametrize(
+    "supervisorctl_path,supervisorctl_conf",
+    [
+        ("supervisorctl", None),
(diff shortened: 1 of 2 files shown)


philpep:
Merged, thanks!

Issue #602, opened 2021-02-20
Ansible hostpattern for ansible connection


ITD27M01:
Hi,

Is there any way to pass the ansible hostpattern to testinfra_hosts? I have dynamic inventory which provides all the hosts from Prod and Dev environments. I'm working with them like:

```
ansible-playbook play.yaml --limit Prod
ansible-playbook play.yaml --limit Dev
```

But both environments have the same groups (db, web, app). Can I pass host pattern like 'db:&Prod' to testinfra_hosts ?

https://docs.ansible.com/ansible/latest/user_guide/intro_patterns.html

philpep:
Hi, testinfra will match hosts and/or groups but no patterns yet. I guess this should be feasible in the future.

ITD27M01:
Hi @philpep 

As an example, you can consider the corresponding lookup plugin - [inventory_hostnames](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/inventory_hostnames_lookup.html). It returns the list of hostnames which can be used as a source for testinfra_hosts: https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/lookup/inventory_hostnames.py

Is there any reason to use external call to ansible-inventory cmd?

ITD27M01:
A've played with this:

```
> python
Python 3.8.5 (default, Sep  4 2020, 02:22:02) 
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from ansible.parsing.dataloader import DataLoader
>>> from ansible.inventory.manager import InventoryManager
>>> loader = DataLoader()
>>> inventory = InventoryManager(loader=loader, sources='inventory')
>>> inventory.get_hosts(pattern='db:&Prod')
>>> ['my', 'hosts']
```

Inventory sources can be retrieved from config or from cmdline.

philpep:
@ITD27M01 thanks for your inputs !

I think we must avoid to use the "API" because we do this in the past and caused us a lot of issues (tl;dr; ansible API isn't stable and practical).
I didn't manage to run `inventory_hostnames` module with the command line but I just saw `ansible` has a `--list-hosts` option that output hosts matching pattern:

``ansible -m noop --list-hosts <pattern>``

If some of you have some time to spend on it that would be an awesome contribution :)

amarao:
I played a bit with `--list-hosts`, and it works.

patterns:
[...]

PR #609, opened 2021-04-07
Add a magical function testinfra_hosts_update to update host list in runtime

amarao:
This adds a support for optional magical function `testinfra_hosts_update` to each module, which allows to filter host list after it was created by `testinfra.get_hosts` before doing an actual test parametrization. Practially, it's allow to filter/update/mangle host list after it was become a list of concrete hosts.

The issue it tries to solve is that `testinfra_hosts` when used in tests for molecule/ansible environment usually contains group names (as host names are volatile in many molecule usecases), and some tests may requires to be "run_once" for a given group.

The practical scenario is like this (mons is a groups for Ceph monitors):
```python
testinfra_hosts = ["ansible://mons?force_ansible=True"]

def testinfra_hosts_update(host_list):
    return testinfra[host_list[0]]

def test_cluster_OK(host):
   assert 'HEALTH_OK' in host.check_output('ceph -s').stdout
```

### DIFF
--- a/testinfra/plugin.py
+++ b/testinfra/plugin.py
@@ -117,6 +117,10 @@ def pytest_generate_tests(metafunc):
             ansible_inventory=metafunc.config.option.ansible_inventory,
             force_ansible=metafunc.config.option.force_ansible,
         )
+        if hasattr(metafunc.module, "testinfra_hosts_update"):
+            if not callable(metafunc.module.testinfra_update.callable):
+                pytest.fail("testinfra_hosts_update must be a function")
+            params = metafunc.module.testinfra_hosts_update(params)
         params = sorted(params, key=lambda x: x.backend.get_pytest_id())
         ids = [e.backend.get_pytest_id() for e in params]
         metafunc.parametrize(
--- a/doc/source/invocation.rst
+++ b/doc/source/invocation.rst
@@ -35,6 +35,12 @@ You can also set hosts per test module::
         [....]
 
 
+If you need to filter or update generated host list (f.e. after ["ansible://somegroup"]),
+you can use testinfra_hosts_update function:
+
+    def testinfra_hosts_update(host_list):
+        return [host_list[0]]
+
 
 Parallel execution
 ~~~~~~~~~~~~~~~~~~


testinfra/plugin.py
```diff
@@ -117,6 +117,10 @@ def pytest_generate_tests(metafunc):
             ansible_inventory=metafunc.config.option.ansible_inventory,
             force_ansible=metafunc.config.option.force_ansible,
         )
+        if hasattr(metafunc.module, "testinfra_hosts_update"):
+            if not callable(metafunc.module.testinfra_update.callable):
+                pytest.fail("testinfra_hosts_update must be a function")
+        params = metafunc.module.testinfra_hosts_update(params)
```
philpep, line 123: This line should also be under the if no ?

testinfra/plugin.py
```diff
@@ -117,6 +117,10 @@ def pytest_generate_tests(metafunc):
             ansible_inventory=metafunc.config.option.ansible_inventory,
             force_ansible=metafunc.config.option.force_ansible,
         )
+        if hasattr(metafunc.module, "testinfra_hosts_update"):
+            if not callable(metafunc.module.testinfra_update.callable):
```
philpep, line 121: Shouldn't this be `callable(metafunc.module.testinfra_hosts_update)` ?

philpep (commented):
Hi thanks for this. But I think it's a bit too magical :)

Maybe It's better to handle ansible patterns no ? See #602

testinfra/plugin.py
amarao: oops. Thank you! I've cleanup code and missed that.

testinfra/plugin.py
amarao: I fixed bad offset from comment above, now it looks like this:
```python
        if hasattr(metafunc.module, "testinfra_hosts_update"):
            if not callable(metafunc.module.testinfra_update.callable):
                pytest.fail("testinfra_hosts_update must be a function")
            params = metafunc.module.testinfra_hosts_update(params)
```
The logic is that we complains if there is something called `testinfra_hosts_update` but it's not callable.

amarao:
> Hi thanks for this. But I think it's a bit too magical :)
> 
> Maybe It's better to handle ansible patterns no ? See #602

I thought about it. I really would like ability to specify the `hosts-style` host list (something like that: `ansible://mons[0],mgrs,!foobar`, which says 'host #1 from monitors, all mgrs, but not the stuff in foobar group or named foobar'), but I suspect it's hard to implement.

Actually some kind of hook to process the hostlist after it was parsed and processed by Ansible is really welcomed. Ansible inventory is really tricky thing (given the molecule control over inventory), so ability to tweak results are important, I think.

The magicalily of `testinfra_hosts_update` is on pair with `testinfra_hosts`, I think. Better naming may be a good idea, though.

philpep:
@amarao I added some comment on #602 , seems `ansible --list-hosts <pattern>` resolve hosts from the pattern. In any cases I think we need to implement this. And maybe later on we could add such hook if needed.

PR #612, opened 2021-04-21
Systemctl show

romainx:
Hello,

Implementation of `systemctl show` to retrieve the properties of the service (unit) -- implemented only for `systemd`.

Return service properties as a `dict`.
The properties retrieved can be limited to a defined list.

```python
>>> ntp = host.service("ntp")
>>> ntp.show(["FragmentPath", "MainPID"])
{'MainPID': '0', 'FragmentPath': '/lib/systemd/system/ntp.service'}
>>> ntp.show("FragmentPath")
{'FragmentPath': '/lib/systemd/system/ntp.service'}
>>> ntp.show()
{'Type': 'forking', 'Restart': 'no', 'NotifyAccess': 'none', ...
```

Best

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -163,6 +163,19 @@ def test_service(host, name, running, enabled):
     service = host.service(name)
     assert service.is_running == running
     assert service.is_enabled == enabled
+    # check with a list
+    props = service.show(["UnitFileState", "ActiveState"])
+    assert props.get("UnitFileState") == "enabled" if enabled else "disabled"
+    assert props.get("ActiveState") in ["active"] if running else ["failed", "inactive"]
+    # check with a single property
+    props = service.show("UnitFileState")
+    assert props.get("UnitFileState") == "enabled" if enabled else "disabled"
+    # check without any property
+    props = service.show()
+    assert props.get("UnitFileState") == "enabled" if enabled else "disabled"
+    # check with something that does not exist
+    props = service.show("DoesNotExist")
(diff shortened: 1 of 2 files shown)


philpep:
Thanks for this! It looks useful.

But the API seems a bit weird to me for instance ```assert ntp.show(["MainPID"])["MainPID"] == "0"``` is a bit repetitive. Also ``show`` isn't clear to me.

I'd suggest this API instead:

```
service = host.service("ntp")
assert service.systemd_properties["MainPID"] == "0"
assert service.systemd_properties["FragmentPath"] == "/lib/systemd/system/ntp.service"
```

``systemd_properties`` would be a ``cached_property`` or ``systemctl show`` dict. I think we can avoid implementing ``--property`` since it doesn't cost that much to systemd to output the whole properties.

What do you think ?

romainx:
@philpep I see what you suggest. It's done in the same way with `cached_property` elsewhere in the codebase. I will change the implementation, you're right it will be much clearer. `show` is versatile, it can be used in different ways so using the naming `systemd_properties` will help clarify its purpose in this context.

> If no argument is specified, properties of the manager will be shown. If a unit name is specified, properties of the unit are shown, and if a job ID is specified, properties of the job are shown.

romainx:
@philpep I think I have implement the changes. Please could you confirm if it's OK for you?
You were right, it's much better this way 😅.

Note: I've also fixed the upstream of the arch linux image that was not found.

philpep:
Merged, thanks!

PR #615, opened 2021-06-02
Interface names and default

### LINKED ISSUE #613: Add an interface list or the default interface
Hi there

I'm doing tests like that

```
def test_portal(host):
    ip = host.interface("eth0").addresses[0]
    get = requests.get("http://" + ip + ":3000/#")
    assert get.status_code == 200
    assert "Portal" in get.text, get.text
```

My issue is, the interface name might change. eth0 for docker container, tap0 for podman's, ens6 for kvm/libvirt based stuff. I'ld like to come up with a generic approach. Could you add a `host.interfaces` list, dict or `host.default_interface` something ?

romainx:
Hello,

First try to resolve #613.
@rgarrigue not sure if it matches your expectations.

```python
# all interfaces
host.interface.names()
# ['lo', 'tunl0', 'ip6tnl0', 'eth0']

# default interface
default_itf = host.interface.default()
default_itf.name
# 'eth0'
default_itf.addresses
# ['172.17.0.3']
```

It's a work in progress. Open to suggestions and improvements either in the naming and in the implementation.

Best

### DIFF
--- a/testinfra/modules/interface.py
+++ b/testinfra/modules/interface.py
@@ -49,6 +49,24 @@ class Interface(Module):
             return BSDInterface
         raise NotImplementedError
 
+    @classmethod
+    def names(cls):
+        """Return the names of all the interfaces.
+        
+        >>> host.interface.names()
+        ['lo', 'tunl0', 'ip6tnl0', 'eth0']
+        """
+        raise NotImplementedError
+
+    @classmethod
+    def default(cls):
+        """Return the name of the default interface.
+        
+        >>> host.interface.default()
+        'eth0'
+        """
+        raise NotImplementedError
+
 
 class LinuxInterface(Interface):
     @cached_property
@@ -72,6 +90,23 @@ class LinuxInterface(Interface):
             if splitted and splitted[0] in ("inet", "inet6"):
                 addrs.append(splitted[1].split("/", 1)[0])
         return addrs
+
+    @classmethod
+    def default(cls):
+        out = cls.check_output("%s route ls", cls(None)._ip)
+        for line in out.splitlines():
+            if "default" in line:
+                out = line.strip().rsplit(" ", 1)[-1]
+        return out
+
+    @classmethod
+    def names(cls):
+        # -o is to tell the ip command to return 1 line per interface
+        out = cls.check_output("%s -o link show", cls(None)._ip)
+        interfaces = []
+        for line in out.splitlines():
+            interfaces.append(line.strip().split(": ", 2)[1].split("@", 1)[0])
+        return interfaces
 
 
 class BSDInterface(Interface):
--- a/testinfra/modules/interface.py
+++ b/testinfra/modules/interface.py
@@ -70,7 +70,7 @@ class Interface(Module):
 
     @classmethod
     def default(cls, family=None):
-        """Return the default interface.
+        """Return the interface used for the default route.
 
         >>> host.interface.default()
         <interface eth0>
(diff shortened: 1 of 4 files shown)


rgarrigue:
Looks good to me. While at it maybe you could add default_ipv4 & default_ipv6 ? 

My own usage is to test ansible role result from outside, to make sure firewall rules are OK, so it's all about IP.

```
get = requests.get("http://" + _get_ip(host) + ":3000/version.txt")
assert get.status_code == 200
```

romainx:
@rgarrigue thanks for the feedback. 
I've not answered exactly to your point, however it made me think that `host.interface.default()` should return an `Interface` that can be used directly (so it's now a constructor).

romainx:
I will work to introduce the _protocol family_ into the module.

> protocol family identifier: inet, inet6 or link ,enforce the protocol family to use. If the option is not present, the protocol family is guessed from other arguments. If the rest of the command line does not give enough information to guess the family, ip falls back to the default one, usually inet or any. link is a special family identifier meaning that no networking protocol is involved. ([source](https://linux.die.net/man/8/ip))

romainx:
@philpep I've made some progress on the implementation. Could you have a look 👀? 
If it's going on the right direction for you I will implement the protocol family in `BSDInterface`.
Thanks.

tox.ini
```diff
@@ -10,6 +10,7 @@ description = Runs unittests
 deps=
     -rtest-requirements.txt
 commands=
+    {envpython} -m pip install types-paramiko
```
philpep, line 13: This can go to ``deps``

testinfra/modules/interface.py
```diff
@@ -49,30 +59,74 @@ def get_module_class(cls, host):
...
+
 
 class LinuxInterface(Interface):
     @cached_property
-    def _ip(self):
-        return self.find_command("ip")
+    def _ip(self, include_family=True):
+        ip_cmd = self.find_command("ip")
+        if include_family and self.family is not None:
+            ip_cmd = f"{ip_cmd} -f {self.family}"
```
philpep, line 91: The ``include_family`` parameter is always True and will always be True since you can't call a ``property`` with parameters. So you can safely drop this variable ;)

testinfra/modules/interface.py
```diff
@@ -49,30 +59,74 @@ def get_module_class(cls, host):
...
+
+        >>> host.interface.names()
+        ['lo', 'tunl0', 'ip6tnl0', 'eth0']
+        """
+        raise NotImplementedError
+
+    @classmethod
+    def default(cls, family=None):
+        """Return the default interface.
+
```
philpep, line 74: Maybe explain here "interface used for the default route".

philpep (changes requested):
Hi thanks for this!

I made few comments, otherwise this looks good to me.

Note ``ip`` has a ``--json`` parameter to output json, but sadly this isn't supported yet in buster & centos 7, so I guess our best choice is to wait few years ;)

tox.ini
romainx: Sure got it, it can be simply added to the `test-requirements.txt`.

testinfra/modules/interface.py
romainx: Indeed! 😓

testinfra/modules/interface.py
romainx: Ok

romainx:
@philpep thanks for the review, I think everything is OK for the `LinuxInterface`. Do you think is worth implementing it in the `BSDInterface`?

philpep:
Merged, thanks!

I don't know if people use testinfra with bsd, but if you have some time to spend on it feel free ;)

rgarrigue:
👏🏿 

Which version will it be released in & when please ?

philpep:
@rgarrigue it's released as 6.4.0

PR #624, opened 2021-09-18
Add more typing

philpep:


ssbarnea:
@philpep Do not lose hope, we all want type! ;)

PR #625, opened 2021-09-26
expr_form is depricated in salt backend client, replaced with tgt_type

sealeg:
Use of expr_form is depricated and no longer works in the latest salt client version.


philpep:
Merged, thanks!

PR #628, opened 2021-10-14
Use RpmPackage on CentOS and Fedora and not only on CentOS

dcermak:
On Fedora you can also install dpkg, which will cause testinfra to think that
the package manager is actually DPKG and not RPM. Thus we instead use the
system_info.distribution value to prefer rpm on Fedora


philpep:
Merged, thanks!

PR #629, opened 2021-11-01
Fix the CI

dcermak:
Two tiny changes to make the CI green again:
- bump release of the test package in the Alpine image
- force a more recent version of the hacking plugin


philpep:
Merged, thanks!

PR #630, opened 2021-11-01
Support older busybox

dcermak:
Testinfra uses `readlink -f` and `command -v` which are both not supported by older versions of busybox. However this can be worked around by using `realpath` and `which`, respectively.

### DIFF
--- a/testinfra/host.py
+++ b/testinfra/host.py
@@ -29,7 +29,11 @@ class Host:
 
     def exists(self, command):
         """Return True if given command exist in $PATH"""
-        return self.run_expect([0, 1, 127], "command -v %s", command).rc == 0
+        rc = self.run_expect([0, 1, 127], "command -v %s", command).rc
+        if rc == 127:
+            return self.run_expect([0, 1], "which %s", command).rc == 0
+        else:
+            return rc == 0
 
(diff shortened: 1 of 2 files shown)


testinfra/modules/file.py
```diff
@@ -61,7 +61,7 @@ def linked_to(self):
         >>> host.file("/var/lock").linked_to
         '/run/lock'
         """
-        return self.check_output("readlink -f %s", self.path)
+        return self.check_output("realpath %s", self.path)
```
philpep, line 64: I wonder what the status of availability of realpath and readlink binaries in various linux/bsd distros...

Maybe we could try one and then another.

testinfra/host.py
```diff
@@ -29,14 +29,21 @@ def __repr__(self):
 
     def exists(self, command):
         """Return True if given command exist in $PATH"""
-        return self.run_expect([0, 1, 127], "command -v %s", command).rc == 0
+        return (
+            self.run_expect(
+                [0, 1, 127], "command -v %s || which %s", command, command
+            ).rc
+            == 0
+        )
```
philpep, line 37: I think we sould run `which` only if `command` exit with 127 (not 1).

Also the implementation command (command or which) could be in a cached_property on the host object.

philpep:
Merged thanks!

PR #636, opened 2022-01-28
Fix Salt decode error

scornelissen85:
See https://github.com/pytest-dev/pytest-testinfra/pull/566. It's the same PR but it's now updated with latest changes from the master branch.


philpep:
Merged, thanks!

PR #638, opened 2022-01-31
add feature for checking user password min/max days

### LINKED ISSUE #637: Add methods to User module for checking minimum/maximum days for password rotation
Need to verify that user accounts with passwords are rotated on a regular basis. Would be helpful if the User module in testinfra could check the minimum/maximum days for password changes of a user, in addition to the account expiration that's already available ie.

```
$ chage -l myuser
Last password change					: Oct 05, 2021
Password expires					: Jan 03, 2022
Password inactive					: Feb 02, 2022
Account expires						: never
Minimum number of days between password change		: 7
Maximum number of days between password change		: 90
Number of days of warning before password expires	: 7
```

jrbeilke:
resolves #637


philpep:
Merged, thanks!

PR #640, opened 2022-02-09
host.py: Fix "which" fallback in find_command

apatard:
Commit "Fallback to which when command -v fails" has changed things to call which(1) in find_command
but used "self.path" has argument to which(1), leading to:

E       AttributeError: 'Host' object has no attribute 'path'

Update the call to pass "command" as argument.

Signed-off-by: Arnaud Patard <apatard@hupstream.com>


philpep:
Wow, thanks for catching this! Merged, thanks!

Issue #467, opened 2019-07-29
3.0 breaks connections using ansible (part 2)


mrdima:
Version 3.0 (all from .0.0 to .0.5) breaks connections to ansible hosts behind a "jumphost or bastionhost". (#439 Doesn't fix it)
What worked fine in 2.x version now doesn't. For some reason testinfra wants to lookup host dns names locally, but that is in our case impossible. They can only be resolved when using the jumphost.

Consider the following inventory:
```
[all:vars]
ansible_ssh_user           = myuser
ansible_python_interpreter = /opt/pypy/bin/pypy
ansible_ssh_common_args    = "-o StrictHostKeyChecking=no -o ProxyCommand=\"ssh -o StrictHostKeyChecking=no -W %h:%p -q {{ansible_ssh_user}}@{{jumphost}} -p 12345\""
jumphost                   = 123.123.1.1


[jumphost]
123.123.1.1:12345
[nodes]
da0-n-8ca8
da0-n-2bc3
da0-n-1ee6
```
Which basically says (using the common_args) that it needs to reach everything using the jumphost ssh running on port 12345
Then running:
```py.test --color=yes -n 3 -vv --hosts='ansible://nodes' --ansible-inventory /inventory -p no:cacheprovider```

Results in:
```================================================================== test session starts ==================================================================
platform linux -- Python 3.7.3, pytest-5.0.1, py-1.8.0, pluggy-0.12.0 -- /usr/bin/python3.7
rootdir: /tests
plugins: forked-1.0.2, xdist-1.29.0, testinfra-3.0.5
[gw0] linux Python 3.7.3 cwd: /tests
[gw1] linux Python 3.7.3 cwd: /tests
[gw2] linux Python 3.7.3 cwd: /tests
[gw0] Python 3.7.3 (default, May  3 2019, 11:24:39)  -- [GCC 8.3.0]
[gw1] Python 3.7.3 (default, May  3 2019, 11:24:39)  -- [GCC 8.3.0]
[gw2] Python 3.7.3 (default, May  3 2019, 11:24:39)  -- [GCC 8.3.0]
gw0 [3] / gw1 [3] / gw2 [3]
scheduling tests via LoadScheduling

test_t.py::test_update_settings[ansible://da0-n-8ca8] 
test_t.py::test_update_settings[ansible://da0-n-2bc3] 
test_t.py::test_update_settings[ansible://da0-n-1ee6] 
[...]

PR #642, opened 2022-02-15
add ssh_extra_args option

dalvizu:
This is follow up to https://github.com/pytest-dev/pytest-testinfra/commit/b749619fbd28a90c3e323fa66b2329aff60facf4 and seems related to https://github.com/pytest-dev/pytest-testinfra/issues/467 but does not solve it:

Expose the ssh_extra_options as a pytest CLI option. This is already accepted as a kwarg to the ssh backend. My use-case was to disable SSH host key checking, I was able to work-around it via a configfile but this seemed like a nice feeature to expose up as its own option, and comments in #467 seem to indicate this was the intent.


philpep:
Merged, thanks!

PR #655, opened 2022-06-16
Adding AlmaLinux and Rocky to package.py

srbala:
Adding AlmaLinux and Rocky to package.py for package detection


srbala:
@team the build failure does not seems related to PR, @philpep Please review, comment or merge. This would be great addition CentOS users from past to use `AlmaLinux` of `Rocky Linux`

philpep:
Merged, thanks!

srbala:
@philpep thank you

PR #662, opened 2022-09-12
Improved ssh config support in Paramiko backend

tpmullan:
The `include` statement in the ssh config is now supported when using paramiko as the backend

This is a relatively simple change, but it allows my overly complicated ssh setup to work out of the box.  Please let me know if there is anything I need to change so this can be merged.


philpep:
Merged, thanks!

PR #664, opened 2022-11-04
Add support for Manjaro-Linux, really closes #503

### LINKED ISSUE #503: Add manjaro linux support
it same as an archlinux, but host.system_info.distribution  is manjarolinux instead of arch

jogi-k:
Why treat manjarolinux <=> arch different than 
e.g. debian <=> ubuntu 
or fedora <=>  centos
==> 
Add support for ManjaroLinux via detection of distribution


philpep:
Merged, thanks!

PR #666, opened 2022-11-13
Remove mock from test requirements

dvzrv:
test-requirements.txt:
Remove mock from test requirements as it is not used.


philpep:
Merged, thanks!

PR #667, opened 2022-11-23
Add support for Zoned Block Devices (ZBD) to LinuxBlockDevice.

xsub:
This change is adding Zoned Block Device support to LinuxBlockDevice in BlockDevice Module.

Currently `cat`-ing a `/sysfs/block/<dev name>/queue/<zoned related file>`s is used query a drive for ZBD options.

Added properties "zoned", "zoned_type" and populated LinuxBlockDevice._data with ZBD configuration values.

Added get_zoned_param() method for querying the zoned device (ZBD-related) configuration values.

Configuration values names allowed as parameters in get_zoned_param() queries:
- "zoned_type"
- "chunk_sectors"
- "nr_zones"
- "zone_append_max_bytes" 
- "max_open_zones"
- "max_actives_zones".


xsub:
:enhancement:

philpep:
Hi, sorry but I think this is too specific for a few users which can implement their own solution using `check_output("cat ...")`.

PR #669, opened 2022-12-03
Fix `command -v` compatibility with dash shell

### LINKED ISSUE #668: Host.exists(): which(1) fallback forced incorrectly when dash(1) shell is being used
The `Host.exists()` method uses the following logic:

https://github.com/pytest-dev/pytest-testinfra/blob/5cd690cd311002032cec0d0b797242e51fd9ca61/testinfra/host.py#L30-L36

This assumes that 127 exit status means that `command -v` is not available and a fallback to which(1) is forced. However, this is incorrect and 127 exit status is used e.g. by dash(1) shell to indicate that the requested executable is not available, i.e.:

```
$ sh -c 'command -v foobar ; echo $?'
127
```

This is a valid behavior [according to POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html#tag_20_22_14).

We've noticed because we're working on making which(1) entirely optional on Gentoo systems, and right now testinfra is incorrectly crashing because of missing which(1) when dash is used as `/bin/sh`.

mgorny:
Fix the `command -v` exit status handling to account for dash being `/bin/sh` (this is e.g. the default on Debian).  Unlike bash, dash exits with status 127 if the requested program does not exist -- and testinfra wrongly interpreted that as `command` being unavailable.

Instead of attempting to use `command -v` and `which` in order, peform an additional (cached) `command -v command` invocation to determine whether `command` is available.  Furthermore, assume that `command -v` can return any non-zero status for command not existing, per POSIX.

This fixes exists() and find_command() raising an exception on systems using dash as `/bin/sh` and lacking `which(1)`.

Fixes #668

### DIFF
--- a/testinfra/host.py
+++ b/testinfra/host.py
@@ -14,6 +14,7 @@ import os
 
 import testinfra.backend
 import testinfra.modules
+from testinfra.utils import cached_property
 
 
 class Host:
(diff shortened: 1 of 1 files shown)


philpep:
Merged, thanks for this!

mgorny:
Thank you!

PR #673, opened 2022-12-23
Add basic docu for file.contains

sandzwerg:
When using this today, I struggled to find how this works in the documentation. This hopefully gives some pointers for people how to structure their pattern. I wanted to find out how to search for "term1" or "term2" which in gnu grep (or extended grep in other grep versions) would be "term1\|term2" but didn't include this as this depends too much on the grep used. Feel free to change or ask for improvements. Not fully satisfied yet and did this via web so not sure if flake8 is happy.

### DIFF
--- a/testinfra/modules/file.py
+++ b/testinfra/modules/file.py
@@ -116,6 +116,9 @@ class File(Module):
         raise NotImplementedError
 
     def contains(self, pattern):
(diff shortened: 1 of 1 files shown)


sandzwerg:
It seems Actions only fails because of missing labels which I can't add myself.

philpep:
Merged, thanks!

PR #678, opened 2022-12-27
Add is_restarting in docker module

### LINKED ISSUE #659: `docker` module should have `is_restarting` check
Currently, the `docker` module has `is_running`, which checks the `State.Running` field of the output of `docker inspect`. For containers that are restarting in a loop (take a container whose service exits when starting with a bad config, and the restart policy is set to keep trying), this assertion would succeed:

```
assert host.docker('my_container').is_running
```

But, the container is actually broken because it is in a boot loop:
```
        "State": {
            "Status": "restarting",
            "Running": true, // <-- tested field
            "Paused": false,
            "Restarting": true,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 0,
            "ExitCode": 1,
            "Error": "",
            "StartedAt": "2022-08-15T15:05:19.198434527Z",
            "FinishedAt": "2022-08-15T15:05:19.28670725Z"
        },
```

Therefore, a new method should be added.

[Example implementation:](https://github.com/pytest-dev/pytest-testinfra/blob/master/testinfra/modules/docker.py#L43)

```
@property
    def is_restarting(self):
        return self.inspect()["State"]["Restarting"]
```

MahdiAbbasi95:
Hi,

I've added is_restarting to the docker module regarding issue #659.
Fix #659

### DIFF
--- a/testinfra/modules/docker.py
+++ b/testinfra/modules/docker.py
@@ -42,6 +42,10 @@ class Docker(Module):
         return self.inspect()["State"]["Running"]
 
     @property
+    def is_restarting(self):
+        return self.inspect()["State"]["Restarting"]
+
+    @property
     def id(self):
         return self.inspect()["Id"]
 


MahdiAbbasi95:
@philpep 
Hi,
Could you check this pull request, please?

philpep (approved):

philpep:
Merged, thanks!

PR #679, opened 2023-01-04
Remove PytestDeprecationWarning about config opts

kbaikov:
Pytest has deprecated markers for hooks configuration. See: https://docs.pytest.org/en/latest/deprecations.html#configuring-hook-specs-impls-using-markers

Each envocation of pytest with python warning enabled would produce the following message:
PytestDeprecationWarning: The hookimpl pytest_configure uses old-style configuration options (marks or attributes).

Please use the pytest.hookimpl(trylast=True) decorator instead

This commit fixes that issue.


philpep:
Merged, thanks for this!

PR #681, opened 2023-01-30
Conditional Check Hardlink to Busybox Binary

gurkenhabicht:
Links to Busybox can be hardlinks as well, an example is the [Prometheus node exporter Helm chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-node-exporter). `get_module_class` is not sufficient in this case.
Therefore, to identify a hardlink, property `inode` is implemented as a GNUFile property. The added condition compares Inodes of `/bin/ps` and `/bin/busybox` as well as the simple existance of busybox itself. Some testcases for the hardlink check via `inode` property were created in `test/test_modules.py`.

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -350,6 +350,16 @@ def test_file(host):
     assert f == host.file("/d/f")
     assert not d == f
 
+    host.check_output("ln /d/f /d/h")
+    hardlink = host.file("/d/h")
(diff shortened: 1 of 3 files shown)


philpep (approved):

philpep:
Merged, thanks!

PR #682, opened 2023-01-30
delegate_to fixture to allow delegation from within test

amarao:
delegate_to is a fixture-function which allow to create new Host objects inside test.

Rationale: There are services where tests, aiming to one host, need to query something or test something on a different host (like a cluster primary, centralized monitoring, etc).

Ansible has a nice idea of task delegation (add `delegate_to: foobar`, and this task is run on another host with the same variables as original host). Adding `delegate_to` into pytest allows to use the same pattern for tests.

Examples:

```python

    @pytest.mark.testinfra_hosts("ansible://foo")
    def test_bar_configured_with_foo(host, delegate_to):
        foo_config = host.check_output("foo --get-foo")
        bar_host = delegate_to("ansible://bar")
        assert bar_host.check_output(f"bar --check {foo_config}") == "OK"
```

```python

    # this test is not dependent on testinfra_hosts value
    # because it's not using 'host' fixture.
    def test_skip_if_empty(delegate_to):
        bar_host = delegate_to("ansible://bar", skip_empty=True)
            assert bar_host.check_output(f"foobar")
```

### DIFF
--- a/doc/source/examples.rst
+++ b/doc/source/examples.rst
@@ -179,3 +179,30 @@ fixture.
     def test_myimage(host):
         # 'host' now binds to the container
         assert host.check_output('myapp -v') == 'Myapp 1.0'
+
+.. _using delegation:
+
+Using delegation
+~~~~~~~~~~~~~~~~
+
+If testinfra is used with Ansible backend, it is possible to delegate
+to another host using `delegate_to` fixture.
+
+.. code-block:: python
+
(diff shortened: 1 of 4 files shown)


philpep:
Hi, I think we need a dedicated fixture for this since this is already working and documented: https://testinfra.readthedocs.io/en/latest/api.html

PR #688, opened 2023-03-08
Add File.is_executable

CarstenGrohmann:
Hi,

this PR contains `File.is_executable` and some additional docstrings.

This simplified the check if a script/program is executable

```
f = host.file('/path/to/myscript')
assert f.is_executable
```

instead of
```
f = host.file('/path/to/myscript')
assert f.exists and f.mode == 0o700
```

Regards,
Carsten

### DIFF
--- a/testinfra/modules/file.py
+++ b/testinfra/modules/file.py
@@ -36,10 +36,12 @@ class File(Module):
 
     @property
     def is_file(self):
+        """Test if the path is a regular file"""
         return self.run_test("test -f %s", self.path).rc == 0
 
     @property
     def is_directory(self):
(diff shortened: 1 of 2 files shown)


testinfra/modules/file.py
```diff
@@ -18,6 +18,9 @@
 class File(Module):
     """Test various files attributes"""
 
+    path = None
+    """Pathname"""
+
```
philpep, line 23:
```suggestion
```

philpep (changes requested):

philpep (approved):

PR #690, opened 2023-03-24
Add propery Group.user_list

CarstenGrohmann:
Hi,

this PR contains `Group.user_list` which returns a list of user names that are members of this group.

Regards,
Carsten

### DIFF
--- a/testinfra/modules/group.py
+++ b/testinfra/modules/group.py
@@ -37,5 +37,13 @@ class Group(Module):
     def gid(self):
         return int(self.check_output("getent group %s | cut -d':' -f3", self.name))
 
+    @property
+    def user_list(self):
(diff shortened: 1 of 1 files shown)


testinfra/modules/group.py
```diff
@@ -37,5 +37,13 @@ def exists(self):
     def gid(self):
         return int(self.check_output("getent group %s | cut -d':' -f3", self.name))
 
+    @property
+    def user_list(self):
```
philpep, line 41:
```suggestion
    def members(self):
```

or `users()`

philpep (changes requested):

philpep (approved):

philpep:
Merged, thanks!

PR #695, opened 2023-05-09
Workaround Docker#42442 by using IPv4 SSH only

an-ky:
As described in https://github.com/moby/moby/issues/42442, docker daemon starts to expose a container port on different host ports for IPv4 and IPv6 which breaks the ssh/safe-ssh/ansible backend tests for some containers.
To prevent this behavior and flaky tests, only the first line from the `docker port %s 22` command gets parsed to acquire the mapped SSH port.


philpep (approved):

philpep:
Merged, thanks!

PR #699, opened 2023-05-09
Add bond module 

an-ky:
Add a Bond module to allow testing bond network devices.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -114,6 +114,14 @@ def build_docker_container_fixture(image, scope):
 
         port = check_output("docker port %s 22", docker_id)
         port = int(port.rsplit(":", 1)[-1])
+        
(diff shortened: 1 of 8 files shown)


test/conftest.py
```diff
@@ -122,6 +130,22 @@ def teardown():
...
+    # without intervening with the default network
+    for idx in range(2):
+        cmd = [
+            "docker",
+            "network",
+            "create",
+            "--driver=bridge",
+            f"infratest-net{idx}",
+            f"--subnet=172.19.{idx}.0/24",
+        ]
```
an-ky, line 143: I'm not happy with that, as it leaves the two networks remaining on the system executing the tests. I was not able to find a better place to put it and neither did I find a place to tidy up this completely. Something like a global fixture.

I'm not that familiar with pytest and all the possibilities to setup test environments.

PR #708, opened 2023-06-20
Retry SSH on ConnectionResetError

tacerus:
Hello,

I faced the same issue described in https://github.com/pytest-dev/pytest-testinfra/pull/66#issuecomment-190262287, however https://github.com/pytest-dev/pytest-testinfra/pull/71 did not help, as the exception in my case was not a Paramiko `SSHException`, but a `ConnectionResetError`.

This patch adds the latter exception to be handled equally.


philpep:
Merged, thanks!

tacerus:
Thanks as well! :-)

PR #711, opened 2023-06-27
Query all usernames and group names

CarstenGrohmann:
Add support to query all local as well as all local and all remote user and group names.


testinfra/modules/user.py
```diff
@@ -133,6 +133,34 @@ def expiration_date(self):
...
+    @property
+    def get_all_users(self):
+        """Returns a list of local and remote user names
+
+        >>> host.user().get_all_users
+        ["root", "bin", "daemon", "lp", <...>]
+        """
+        all_users = [
+            line.split(":")[0]
+            for line in self.check_output("getent passwd").splitlines()
```
martinhoyer, line 145: I think we could use `[pwd](https://docs.python.org/3.13/library/pwd.html#module-pwd)`? 
```
import pwd
all_users = [pw.pw_name for pw in pwd.getpwall()]
```

CarstenGrohmann: Unfortunately, this is not possible because testinfra only executes Unix commands on the remote host.

philpep (approved):

philpep:
Merged, thanks!

Issue #619, opened 2021-07-17
testinfra should be a typed library (mypy)


ssbarnea:
For reasons I will not bother to mention, testinfra should be a fully typed library. I am aware that this is a big challenge but the good news is that this can be done gradually, in a progressive manner.

Please CC me on any PR that improve type system, I am more than happy to help getting these in.

PR #717, opened 2023-07-20
Add type annotations for backend/base.py

amaslenn:
To start with #619.

@ssbarnea could you please review?

I don't know how to configure `mypy` to check one module with different rules. If that is possible, we can lint `backend/base.py` more strictly (more changes will be needed).


testinfra/backend/base.py
```diff
@@ -26,25 +27,25 @@
...
-        self._stderr = stderr
-        self.command = command
-        self._backend = backend
+        self.exit_status: int = exit_status
+        self._stdout_bytes: bytes = stdout_bytes
+        self._stderr_bytes: bytes = stderr_bytes
+        self._stdout: Optional[str] = stdout
+        self._stderr: Optional[str] = stderr
+        self.command: bytes = command
+        self._backend: BaseBackend = backend
```
philpep, line 44:
```suggestion
        self.exit_status = exit_status
        self._stdout_bytes = stdout_bytes
        self._stderr_bytes = stderr_bytes
        self._stdout = stdout
        self._stderr = stderr
        self.command = command
        self._backend = backend
```

testinfra/backend/base.py
```diff
@@ -112,12 +113,19 @@ class BaseBackend(metaclass=abc.ABCMeta):
...
+        *args: Any,
+        **kwargs: Any,
+    ):
         self._encoding = None
         self._host = None
         self.hostname = hostname
-        self.sudo = sudo
-        self.sudo_user = sudo_user
+        self.sudo: bool = sudo
+        self.sudo_user: Optional[bool] = sudo_user
```
philpep, line 128:
```suggestion
        self.sudo = sudo
        self.sudo_user = sudo_user
```

philpep (approved):

philpep:
Merged with minor modification (no need to re-type variables in __init__()).

Thanks!

We now have a strict mypy configuration for typed code https://github.com/pytest-dev/pytest-testinfra/blob/main/mypy.ini . So I think we can gradually add types from low-level code up to all code.

However I think some part of the code must be simplified to make adding types easier (refactoring & use some dataclasses).

PR #722, opened 2023-07-26
Make CommandResult a dataclass

amaslenn:
Following discussion for #717, make `CommandResult` a dataclass.

### DIFF
--- a/testinfra/backend/base.py
+++ b/testinfra/backend/base.py
@@ -17,7 +17,7 @@ import logging
 import shlex
 import subprocess
 import urllib.parse
-from typing import TYPE_CHECKING, Any, Optional
+from typing import TYPE_CHECKING, Any, Optional, Union
 
 if TYPE_CHECKING:
     import testinfra.host
@@ -33,6 +33,7 @@ class HostSpec:
     password: Optional[str]
 
 
+@dataclasses.dataclass
 class CommandResult:
     """Object that encapsulates all returned details of the command execution.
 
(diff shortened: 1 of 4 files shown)


testinfra/backend/base.py
```diff
@@ -301,6 +256,36 @@ def encode(self, data: str) -> bytes:
...
+            err_bytes = args[3]
+
+        out_bytes = kwargs.get("stdout_bytes", out_bytes)
+        err_bytes = kwargs.get("stderr_bytes", err_bytes)
+
+        out = kwargs.get("stdout", "")
+        if not out and out_bytes:
+            out = self.decode(out_bytes)
+        elif out and not out_bytes:
+            out_bytes = self.encode(out)
```
philpep, line 274: I think we should keep lazy encoding or decoding, i.e. using properties, because sometime we may want to get binary data which cannot be converted to text.

I think the dataclass should have `_stdout` and `_stdout_bytes` and check at `__init__` that at least one of them are set. Or make subclasses BinaryCommandResult and TextCommandResult.

amaslenn: Why is lazy encoding and decoding are needed? My way of thoughts here was to remove dependency on `Backend`, which looks redundant to me. So why is it needed here?

With `dataclass` implementation we always know that parameters are set (at least as `None`, but that is the task for `mypy` to detect), so this verification is kinda included.

Current solution doesn't break backward compatibility, at least to the extend it is tested now.

Another solution here would be to do similar to what `subprocess.CompletedProcess` does: it is generic and can be either `bytes` of `str`. We can borrow the approach, but that will change the API.

philpep: Lazy encoding/decoding is needed because commands like `run("cat /usr/bin/cat")` will raise an error if trying to decode. Also IIRC either salt and/or ansible might have weird encoding behavior while returning command output in a json field. Lazy encoding/decoding avoid some error, as long as user doesn't explicitly ask for encoding/decoding.

Yes something like this could be fine:

```python
@dataclass
class CommandResult:
    _stdout: AnyStr
    _stderr: AnyStr

    @property
    def stdout(self) -> str:
        if isinstance(self._stdout, bytes):
            return self._stdout.encode(...)
        return self._stdout

    [...]
```

amaslenn: _(took longer than I expected, I was busy)_

Applied changes to keep lazy encoding/decoding.
@philpep do you think it is time to change `result(self, *args: Any, **kwargs: Any)` signature in this PR?

philpep (approved):

philpep:
I modified `result()` to make it simple and changed calls instead. Merged, thanks!

PR #723, opened 2023-08-03
Improve system detection for openSUSE style distributions

dirkmueller:
openSUSE distributions can have rpm, dnf and dpkg installed at the same time, but the system is always maintained via rpm. so explicitly detect it as such. Also don't trust lsb-release too much, /etc/os-release is the preferred method so pick that one first.


philpep:
Merged, thanks!

PR #724, opened 2023-08-04
Ansible: Fix for missing group names in get_variables()

marcandre-larochelle-bell:
The Ansible variable group names contains all groups the host is part of and not only the direct parent groups, this is a fix to reflect how ansible works.

Link to documentation:
- https://docs.ansible.com/ansible/latest/reference_appendices/special_variables.html#term-group_names

### DIFF
--- a/test/test_backends.py
+++ b/test/test_backends.py
@@ -190,7 +190,7 @@ def test_ansible_get_variables():
             "a": "a",
             "e": "f",
             "inventory_hostname": "rockylinux",
-            "group_names": ["ungrouped"],
+            "group_names": ["all", "ungrouped"],
             "groups": groups,
         }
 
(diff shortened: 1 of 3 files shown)


marcandre-larochelle-bell:
Hi, sorry to tag you @philpep, this should be a quick fix, otherwise I need to monkey patch the fix in my code and I'd rather not. 

Basically it didn't list all the groups the way ansible does it (by including all the groups it is part of, not just the direct parents).  This allows to load the ansible variables the way they should be loaded (by loading the top parents, then overriding by loading the children of the top parent, then by the children of the children).

Thanks in advance.

marcandre-larochelle-bell:
@philpep Done!

philpep:
Merged, thanks!

marcandre-larochelle-bell:
Hi, just wondering if you know when there'll be a release with the fix (maybe 9.0.1?), we're currently using a  monkey patch as a workaround on our side, but we'd much prefer not to! Thanks in advance

PR #726, opened 2023-08-23
Fix output of error message

CarstenGrohmann:
`cmp` is already a string. Therefore, the `join()` instruction is not necessary and only inserts additional spaces.

### DIFF
--- a/testinfra/utils/ansible_runner.py
+++ b/testinfra/utils/ansible_runner.py
@@ -383,9 +383,7 @@ class AnsibleRunner:
                     "msg": "Skipped. You might want to try check=False",
                 }
             if not files:
-                raise RuntimeError(
-                    "Error while running {}: {}".format(" ".join(cmd), out)
-                )
+                raise RuntimeError(f"{out}")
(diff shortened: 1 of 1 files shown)


testinfra/utils/ansible_runner.py
```diff
@@ -383,9 +383,8 @@ def run_module(
                     "msg": "Skipped. You might want to try check=False",
                 }
             if not files:
-                raise RuntimeError(
-                    "Error while running {}: {}".format(" ".join(cmd), out)
-                )
+                cmd = local.backend.get_command(cmd, *args)
+                raise RuntimeError('Error while running "{}": {}'.format(cmd, out))
```
philpep, line 387: `cmd`is same as `out.cmd` and already displayed in `str(out)` no ?

So I suggest:

```python
raise RuntimeError(f"{out}")
```

CarstenGrohmann: Changed in 5a54e25

philpep:
Merged, thanks!

PR #727, opened 2023-08-23
Use new key to specify license file in setup.cfg

CarstenGrohmann:
to prevent deprecation warning.

Please check next builds that the license file is always included as the new metadata works only for wheels - not sdists.


philpep:
Merged, thanks!

PR #730, opened 2023-08-31
New module: iproute2

costasd:
This changeset introduces a new module, called `iproute2`, based on the `Interface` module, that enables testing of different network parameters through the use of different iproute2 tools, such as the `ip` family of tools or `bridge`.

Right now, the following are supported and exposed through relevant methods:
    * ip-address
    * ip-link
    * ip-route
    * ip-rule
    * ip-vrf
    * ip-tunnel
    * ip-netns
    * bridge vlan
    * bridge link
    * bridge fdb
    * bridge mdb

Furthermore, execution can be optionally contained into a specific family (`inet` or `inet6`) or a particular namespace.

Relevant tests and documentation for each of these methods is included with the module code.


costasd:
:wave: Not sure if you'd prefer an accompanying issue with this PR - happy to if needed - Thanks!

PR #732, opened 2023-09-02
Test if package is not installed

700grm:
Instead of using `run`, `exist `or `find_command ` I propose native check `is_not_installed`, as it is more clear to read:

For example:

```
def test_telnet_is_not_installed(host):
    telnet = host.package("telnet")
    assert telnet.is_not_installed
```


philpep:
You can write:

```python
assert not telnet.is_installed
```

PR #740, opened 2023-10-05
Extend list of valid suffixes for systemd units

CarstenGrohmann:
The old implementation limits the usage of SystemdService.is_valid() to `.service` units. This implementation allows checking all types of unit files with `is_valid()` (`systemd-analyze verify <unitname.suffix>`).


philpep (approved):

philpep:
Merged, thanks!

PR #741, opened 2023-10-09
Add Service.exists

CarstenGrohmann:


philpep (approved):
LGTM, probably easy and worth to add a test for this, can be done as a followup PR.

PR #743, opened 2023-10-29
feat(package): brew support

soof-golan:
# Motivation

Allowing the use of more features of `testinfra` on macOS (`darwin`) hosts.

# Changes

* Added HomebrewPackage

## Notes

* Linux hosts with homebrew installed might also work (though untested). Can easily be limited to only `darwin` hosts, leaving this decision up to maintainers' discretion.


philpep (approved):

philpep:
Merged, thanks!

PR #744, opened 2023-11-10
testinfra/modules/interface: Implement link property

BenoitKnecht:
On Linux, implement `host.interface(iface).link()`, which returns the output of

```shell
ip --json link show $iface
```


philpep (approved):

philpep:
Merged, thanks!

PR #745, opened 2023-11-10
testinfra/modules/blockdevice: Don't fail on stderr

BenoitKnecht:
The `blockdevice` module raises a `RuntimeError` if `blockdev` returns with a non-zero exit code, or if stderr is not empty.

But the latter can happen even when there's no failure at all, for instance if `sshd` has a banner message configured (which gets output to stderr whenever someone connects to the host).


philpep (approved):

philpep:
Merged, thanks!

PR #753, opened 2024-01-21
Draft: Replicate more of the ansible style host matching logic

Code0x58:
This expands on the ansible pattern matching so you can use multiple parts, intersection, negation, and regex patterns, on top of the already implemented globbing.

I'd like to backport this change into an 8.2 release if you'll take it.

### DIFF
--- a/testinfra/utils/ansible_runner.py
+++ b/testinfra/utils/ansible_runner.py
@@ -16,8 +16,9 @@ import functools
 import ipaddress
 import json
 import os
+import re
 import tempfile
-from typing import Any, Callable, Iterator, Optional, Union
+from typing import Any, Callable, Iterator, List, Optional, Set, Union
 
 import testinfra
 import testinfra.host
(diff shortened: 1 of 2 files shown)


testinfra/utils/ansible_runner.py
```diff
@@ -26,6 +27,86 @@
 
 local = testinfra.get_host("local://")
 
+Inventory = dict[str, Any]
```
Code0x58, line 30:
```suggestion
Inventory = Dict[str, Any]
```

testinfra/utils/ansible_runner.py
```diff
@@ -16,8 +16,9 @@
 import ipaddress
 import json
 import os
+import re
 import tempfile
-from typing import Any, Callable, Iterator, Optional, Union
+from typing import Any, Callable, Iterator, List, Optional, Set, Union
```
Code0x58, line 21: This and the suggestion below get this working on lower versions of python, probably replacing the desire for back porting to 8.2
```suggestion
from typing import Any, Callable, Dict, Iterator, List, Optional, Set, Union
```

Code0x58 (commented):
looks ok when used in `testinfra_hosts` on my setup. Should apply the change to easily reduce the minimum working python3 version (tested on 3.8)

testinfra/utils/ansible_runner.py
```diff
@@ -26,6 +27,86 @@
...
+            if intersect is None:
+                intersect = expanded
+            else:
+                intersect &= expanded
+        elif requirement.startswith('!'):
+            negative.update(expand_pattern(requirement[1:], inventory))
+        else:
+            positive.update(expand_pattern(requirement, inventory))
+
+    result = positive
```
Code0x58, line 103: _Comment about something that I don't see as worth changing right now as there hasn't been any response on this yet, and a significant improvement is much better than nothing._

I suspect a perfect copy of the logic from ansible would use something like this here:
```suggestion
    if positive:
        result = positive
    else:
        result = expand_pattern('all')
```
Then you imitate behaviour like `!right` which is equivalent to `all:!right`, but this would also be best with additions to manual testing and adding of tests to this codebase for the confirmed functionality.

Given that adding `all:` to the start of a host pattern/query is equivalent, it's a fine enough workaround. If this change gets merged/released, I'll follow up with this change.

Code0x58:
I'm happy with the improvements on here, so have un-drafted the PR

CarstenGrohmann:
What do you think about renaming `test_ansible.py` to `test_ansible_inventory.py` or similar? That would be more meaningful.

PR #756, opened 2024-02-14
Import future annotations for pytest 6, Python 3.9 compatibility

### LINKED ISSUE #755: Broken compatibility with pytest<7 (pytest.Parser type annotation)
The parser annotation in plugin.py introduced in 10.0 does not work with pytest 6. https://github.com/pytest-dev/pytest-testinfra/blob/dc48cd98f3b1970ecbad63f866a94f8283b79ce4/testinfra/plugin.py#L40
`pytest.Parser` is only available from pytest 7.0.0
```
>>> from testinfra import plugin
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/site-packages/testinfra/plugin.py", line 40, in <module>
    def pytest_addoption(parser: pytest.Parser) -> None:
AttributeError: module 'pytest' has no attribute 'Parser'
```

While specifying minimal pytest version would be a solution, I would rather recommend finding another solution, as RHEL-9/Alma Linux 9/CentOS Stream 9 are running pytest 6 (and Python 3.9).

I'd be happy to help with finding a solution, should you decide to keep pytest 6 compatibility.

martinhoyer:
Makes testinfra v10.0.0 work with Python 3.9 and pytest 6.2.2, which is being used in CentOS-Stream-9-based distributions.
 
Resolves #755.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ packages = find:
 setup_requires =
     setuptools_scm
 install_requires =
-    pytest!=3.0.2
+    pytest>=6
 extras_require =
 
 [options.extras_require]
--- a/testinfra/plugin.py
+++ b/testinfra/plugin.py
@@ -10,6 +10,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
 import logging
 import shutil
 import sys


martinhoyer:
I've successfully ran tests on RHEL-9 (Python 3.9) with install_requires `pytest==6.0.0`.
When changed to `pytest<6`, py tests failed with INTERNALERRORs. Adding therefore a `pytest>=6` requirement to the setup.cfg.

philpep (approved):

philpep:
Sounds reasonable, merged, thanks for this!

Issue #758, opened 2024-04-01
`host.package().is_installed` is returning True incorrectly when the RPM database is corrupted.


narmaku:
### Summary
When calling `host.package().is_installed` in a system using RPMs (in our case, RHEL), and the actual RPM database is corrupted, the command returns True, but the package could not even be queried to know if it's installed or not.

We got a false positive in this test case:

```
[gw3] linux -- Python 3.9.16 /opt/app-root/bin/python3.9
[gw3] linux -- Python 3.9.16 /opt/app-root/bin/python3.9[gw3] linux -- Python 3.9.16 /opt/app-root/bin/python3.9

self = <test_aws.TestsAWS object at 0x7fba46f33c70>
host = <testinfra.host.Host paramiko://<masked>.us-gov-east-1.compute.amazonaws.com>

    @pytest.mark.run_on(['rhel'])
    def test_unwanted_packages_are_not_present(self, host):
        """
        Some pkgs are not required in EC2.
        BugZilla 1888695, 2075815
        """
        unwanted_pkgs = [
            'aic94xx-firmware', 'alsa-firmware', 'alsa-lib', 'alsa-tools-firmware',
            'ivtv-firmware', 'iwl1000-firmware', 'iwl100-firmware', 'iwl105-firmware',
            'iwl135-firmware', 'iwl2000-firmware', 'iwl2030-firmware', 'iwl3160-firmware',
            'iwl3945-firmware', 'iwl4965-firmware', 'iwl5000-firmware', 'iwl5150-firmware',
            'iwl6000-firmware', 'iwl6000g2a-firmware', 'iwl6000g2b-firmware', 'iwl6050-firmware',
            'iwl7260-firmware', 'libertas-sd8686-firmware', 'libertas-sd8787-firmware', 'libertas-usb8388-firmware',
            'firewalld', 'biosdevname', 'plymouth', 'iprutils', 'rng-tools', 'qemu-guest-agent'
        ]
    
        system_release = version.parse(host.system_info.release)
    
        # BugZilla 1888695
        if version.parse('8.3') > system_release >= version.parse('8.0'):
            unwanted_pkgs.remove('rng-tools')
    
        if system_release < version.parse('8.5'):
            unwanted_pkgs.remove('qemu-guest-agent')
    
        if test_lib.is_rhel_sap(host):
            # In RHEL SAP images, alsa-lib is allowed
            unwanted_pkgs.remove('alsa-lib')
    
[...]

PR #759, opened 2024-04-01
Improved RpmPackage is_installed property logic.

narmaku:
This new logic verifies if there is a "not installed" string in the stdout when "rpm -q %s" command returns 1.

As seen in bug #758, if rpm command failed because of other reasons (such as database corruption), the old implementation was returning True, making the end-user think that the package was installed, however it was never queried.

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -65,7 +65,7 @@ def test_rpmdb_corrupted(host):
     with pytest.raises(RuntimeError) as excinfo:
         host.package("zsh").is_installed
     assert (
-        f"Could not check if RPM package 'zsh' is installed. "
+        "Could not check if RPM package 'zsh' is installed. "
         "error: sqlite failure:"
     ) in str(excinfo.value)
 
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -58,19 +58,16 @@ def test_held_package(host):
     assert python.is_installed
     assert python.version.startswith("3.11.")
 
+@pytest.mark.destructive
 @pytest.mark.testinfra_hosts("docker://rockylinux9")
 def test_rpmdb_corrupted(host):
-    host.check_output("mv /var/lib/rpm/rpmdb.sqlite /var/lib/rpm/rpmdb.sqlite.bck")
-    try:
-        host.check_output("dd if=/dev/zero of=/var/lib/rpm/rpmdb.sqlite bs=1024 count=1")
-        with pytest.raises(RuntimeError) as excinfo:
-            host.package("zsh").is_installed
-        assert (
-            f"Could not check if RPM package 'zsh' is installed. "
-            "error: sqlite failure:"
-        ) in str(excinfo.value)
-    finally:
-        host.check_output("mv /var/lib/rpm/rpmdb.sqlite.bck /var/lib/rpm/rpmdb.sqlite")
+    host.check_output("dd if=/dev/zero of=/var/lib/rpm/rpmdb.sqlite bs=1024 count=1")
+    with pytest.raises(RuntimeError) as excinfo:
+        host.package("zsh").is_installed
+    assert (
+        f"Could not check if RPM package 'zsh' is installed. "
+        "error: sqlite failure:"
+    ) in str(excinfo.value)
 
 @pytest.mark.testinfra_hosts("docker://rockylinux9")
 def test_non_default_package_tool(host):
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -58,6 +58,19 @@ def test_held_package(host):
     assert python.is_installed
     assert python.version.startswith("3.11.")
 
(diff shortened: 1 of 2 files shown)


narmaku:
@CarstenGrohmann , @philpep since you are the most active collaborators/maintainers, would you mind checking this quick PR? Thanks!

martinhoyer:
lgtm, I'd just use f-strings, as testinfra requires Python>=3.9.  
I'm also interested in learning about review process here, looks like there is quite a number of PRs without a response.

CarstenGrohmann:
LGTM, also. Please add a small test case.

testinfra/modules/package.py
```diff
@@ -165,7 +165,16 @@ def version(self):
 class RpmPackage(Package):
     @property
     def is_installed(self):
-        return self.run_test("rpm -q %s", self.name).rc == 0
+        result = self.run("rpm -q %s", self.name)
```
philpep, line 168: You can keep using `run_test` so in case of exit code other than 0 or 1 this will raise.

testinfra/modules/package.py
```diff
@@ -165,7 +165,16 @@ def version(self):
...
+        if result.succeeded:
+            return True
+
+        if "not installed" in result.stdout:
```
philpep, line 172: Doesn't rpm support many language depending on LANG/LC_ALL environment variables ? In which case this might not work.

philpep (changes requested):

testinfra/modules/package.py
narmaku: I cannot capture the `stdout` with `run_test`.
In the case exit code will be other than 1 or 0 it will throw anyways, as a `RuntimeError` exception.

testinfra/modules/package.py
narmaku: Good point. I will add an additional check here.

testinfra/modules/package.py
```diff
@@ -165,7 +165,14 @@ def version(self):
 class RpmPackage(Package):
     @property
     def is_installed(self):
-        return self.run_test("rpm -q %s", self.name).rc == 0
+        result = self.run("rpm -q %s", self.name)
+        if result.succeeded:
+            return True
+
+        if "not installed" in result.stdout or not result.stdout.startswith(self.name):
```
narmaku, line 172: We can be pickier here and make the `stdout` match a regex, just to make sure there is actually a package with its version, etc. @philpep  What do you think?

test/test_modules.py
```diff
@@ -58,6 +58,19 @@ def test_held_package(host):
...
+@pytest.mark.testinfra_hosts("docker://rockylinux9")
+def test_rpm_package_broken_db(host):
+    with host.sudo():
+        # Corrupt RPM database so that we make RpmPackage.is_installed throw an exception
+        host.run_test('rpm --rebuilddb')
+        host.run_test('dd if=/dev/zero of=$(rpm --dbpath)/Packages bs=1024 count=1024 seek=1024')
+    with pytest.raises(RuntimeError) as excinfo:
+        host.package("zsh").is_installed
+    assert (
+        "DB_PAGE_NOT_FOUND"
```
narmaku, line 71: In the end it seems the string to look at is `DB_PAGE_NOTFOUND`.
We can also check for `-30986` error code which corresponds to the DB_PAGE_NOTFOUND error.

CarstenGrohmann:
I suggest to simplify the test as a corrupt database is not necessary:

```
$ rpm -q not_installed_pkg
package not_installed_pkg is not installed
$ echo $?
```

testinfra/modules/package.py
```diff
@@ -165,7 +165,13 @@ def version(self):
 class RpmPackage(Package):
     @property
     def is_installed(self):
-        return self.run_test("rpm -q %s", self.name).rc == 0
+        result = self.run("rpm -q --quiet %s 2>&1", self.name)
+        if result.succeeded:
+            return True
+        elif result.failed and result.stdout == '':
```
narmaku, line 171: FYI: If return code is `1` but there is no output/error, it means the package is not installed (for that we used `--quiet` option).

testinfra/modules/package.py
```diff
@@ -165,7 +165,13 @@ def version(self):
...
+            return False
+        else:
+            raise RuntimeError(f"Could not check if RPM package '{self.name}' is installed. {result.stdout}")
```
narmaku, line 174: FYI: If return code is `1` and there is `stdout`, that means there was an unexpected error, so we throw a `RuntimeError` here. (as we use `--quiet` there should not be any output when failing).

philpep (approved):
I modified the test with a corrupted rpmdb  and switched back to use run_test().

philpep:
Merged, thanks!

PR #760, opened 2024-04-09
Extend LinuxBlockDevice to use information from lsblk and add method to return all block devices on the system

fsuba:
The purpose of this PR is to give us more information about the block devices on a system. lsblk is part of the util-linux package, the same package as blockdev. With the information from lsblk, we can easily return all the block devices found on a system.

### DIFF
--- a/testinfra/modules/blockdevice.py
+++ b/testinfra/modules/blockdevice.py
@@ -11,6 +11,7 @@
 # limitations under the License.
 
 import functools
+import json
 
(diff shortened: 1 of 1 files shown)


martinhoyer (approved):
I think this is pretty cool. `lsblk -J` option is available even on older versions fwiw.  
It would be nice to have some feedback from the maintainers here :) Does it fit-in to testinfra codebase as is, or is there something that needs to be changed?

PR #761, opened 2024-04-20
Add py.typed marker

pjonsson:
Without the py.typed marker, mypy will complain
for users of testinfra:

file.py:2: error: Skipping analyzing "testinfra.host": module is installed, but missing library stubs or py.typed marker  [import-untyped]
file.py:2: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports

Refs #619


philpep:
Hi, I think we should have 100% typed code before doing this. Otherwise this could be confusing for users.

PR #764, opened 2024-05-07
Remove crypt lib from testing

martinhoyer:
crypt module is being deprecated in Python 3.13 (as well as spwd). It is being used for checking host.user().password. Replacing the crypt.crypt() with hashlib or passlib would not be trivial, due to generated salt, format of shadow, etc.

### DIFF
--- a/test/test_modules.py
+++ b/test/test_modules.py
@@ -10,7 +10,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import crypt
 import datetime
 import os
 import re
@@ -257,7 +256,8 @@ def test_nonexistent_user(host):
 def test_current_user(host):
     assert host.user().name == "root"
     pw = host.user().password
-    assert crypt.crypt("foo", pw) == pw
+    assert pw.startswith("$")
+    assert len(pw) == 73
 
 
 def test_group(host):


martinhoyer:
as for the CI failure, the new version of Salt requires timelib and that does fails to build. Installing gcc or locking salt version could help?

martinhoyer:
Any chance someone could take look please?

philpep (approved):

philpep (changes requested):
Looks test is failing. Even with passlib.hosts.linux_context.verify(). Maybe hash algorithm is not handled by passlib ?

martinhoyer:
> Looks test is failing. Even with passlib.hosts.linux_context.verify(). Maybe hash algorithm is not handled by passlib ?

Thanks for looking into it. Checking what's wrong.

martinhoyer:
ok, turns out it's not as easy as I thought. `crypt.crypt()` does not have a replacement in Python and spwd was also deprecated in 3.13. How would you like to proceed?

philpep:
> ok, turns out it's not as easy as I thought. `crypt.crypt()` does not have a replacement in Python and spwd was also deprecated in 3.13. How would you like to proceed?

I think we can just assert `host.user().password` startswith `$` and eventually has some expected `len()`. IIRC this should be constant, whereas encrypted password value might change between docker builds due do some random hash.

philpep (approved):

PR #767, opened 2024-05-26
Fix service.exists

### LINKED ISSUE #768: service.exists with systemd fails
On a systemd with systemd using SystemdService class, `service.exists` property fails with

````
 AssertionError: Unexpected exit code 2 for CommandResult(backend=<testinfra.backend.docker.DockerBackend object at 0x7f23a47a8520>, exit_status=2, command=b'systemctl list-unit-files | grep -q"^avahi-daemon"', _stdout=b'', _stderr=b"grep: invalid option -- '^'\nUsage: grep [OPTION]... PATTERNS [FILE]...\nTry 'grep --help' for more information.\n")
assert 2 in [0, 1]
+  where 2 = CommandResult(backend=<testinfra.backend.docker.DockerBackend object at 0x7f23a47a8520>, exit_status=2, command=b'syst...=b"grep: invalid option -- '^'\nUsage: grep [OPTION]... PATTERNS [FILE]...\nTry 'grep --help' for more information.\n").rc
````

sblaisot:
service.exists fail with systemd because there was a typo in the underlying command, missing a space between grep option `-q` and service name.

This PR fixes that problem

This fixes #768

### DIFF
--- a/testinfra/modules/service.py
+++ b/testinfra/modules/service.py
@@ -176,7 +176,7 @@ class SystemdService(SysvService):
 
     @property
     def exists(self):
-        cmd = self.run_test('systemctl list-unit-files | grep -q"^%s"', self.name)
+        cmd = self.run_test('systemctl list-unit-files | grep -q "^%s"', self.name)
         return cmd.rc == 0
 
     @property


martinhoyer (commented):
Thanks for the fix.  
I wonder what the `^` is for. Isn't it redundant?

sblaisot:
> I wonder what the `^` is for. Isn't it redundant?

The `^` in the beginning of a regex means "beginning with" (start of line)

I only corrected the missing space between `-q` option and the searched regex.

martinhoyer:
> > I wonder what the `^` is for. Isn't it redundant?
> 
> The `^` in the beginning of a regex means "beginning with" (start of line)
> 

Ah, I always get confused by the % substitution :)

> I only corrected the missing space between `-q` option and the searched regex.

Of course, I was just curious.

martinhoyer (approved):

philpep (approved):

philpep:
> > > I wonder what the `^` is for. Isn't it redundant?
> > I only corrected the missing space between `-q` option and the searched regex.
> 
> Of course, I was just curious.

I think this is to match service with or without the ".service" suffix. Maybe we could think about a better way of doing this because service `s` (which obvisouly doesn't exist) will match `ssh.service` and reported as existing.

philpep:
Merged, thanks!

PR #773, opened 2024-07-02
Fix systemctl is active

arossert:
A fix for this issue
https://github.com/pytest-dev/pytest-testinfra/issues/748

### DIFF
--- a/testinfra/modules/service.py
+++ b/testinfra/modules/service.py
@@ -181,7 +181,13 @@ class SystemdService(SysvService):
 
     @property
     def is_running(self):
-        out = self.run_expect([0, 1, 3], "systemctl is-active %s", self.name)
+        # based on https://man7.org/linux/man-pages/man1/systemctl.1.html
+        # 0: program running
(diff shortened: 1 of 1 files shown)


pfuntner:
I'd like to see this fixed.  It's affecting a project with which I'm involved at work (#775).  I suppose I'll have to do some lame workarounds. :-(

martinhoyer:
@philpep Hi Philippe, is this project looking for maintainers/reviewers? Seems like the open PRs are a bit stuck..

permanentdaylight:
@martinhoyer 

This also applies to `is_enabled`, might be a good idea to get a fix in for that too.

ltog:
I am affected by this too. I checked the changes and they look fine to me.

philpep (approved):
I still think previous behavior (AssertionError) was fine. But I'm also fine with the new proposed behavior. so LGTM.

philpep:
Merged, thanks!

permanentdaylight:
Thanks @philpep

Issue #775, opened 2024-09-12
`host.service('test').is_running` is throwing _Unexpected exit code 4_ exception on Ubuntu 24


pfuntner:
`systemctl is-active FOOBAR` (where FOOBAR is a non-existent service) is behaving differently on Ubuntu 24.  It has exit code 4 rather than exit code 3 on releases such as Ubuntu 20 and Ubuntu 22.

```
$ python3 -m pip list | grep -P pytest\|testinfra
pytest           8.3.3
pytest-testinfra 10.1.1
testinfra        6.0.0
$ cat ~/tmp/test_service.py
def test_service1(host):
  print(f'\ntest_service1: {host.service("test").is_running}')

def test_service2(host):
  running = host.run_expect([0, 1, 3, 4, 8], 'systemctl is-active test').rc == 0
  print(f'\ntest_service2: {running}')
$ pytest -s --connection ssh --hosts gcp-ubuntu2404 ~/tmp/test_service.py
/home/jpfuntne/third-party/pytest/lib/python3.11/site-packages/_testinfra_renamed.py:5: DeprecationWarning: testinfra package has been renamed to pytest-testinfra. Please `pip install pytest-testinfra` and `pip uninstall testinfra` and update your package requirements to avoid this message
  warnings.warn((
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.11.0rc1, pytest-8.3.3, pluggy-1.5.0
rootdir: /home/jpfuntne
plugins: testinfra-10.1.1, testinfra-6.0.0
collected 2 items

../tmp/test_service.py F
test_service2: False
.

========================================================================================== FAILURES ==========================================================================================
____________________________________________________________________________ test_service1[ssh://gcp-ubuntu2404] _____________________________________________________________________________

host = <testinfra.host.Host ssh://gcp-ubuntu2404>

    def test_service1(host):
>     print(f'\ntest_service1: {host.service("test").is_running}')

../tmp/test_service.py:2:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[...]

PR #776, opened 2024-09-13
Adding additional potential systemctl exit status for Ubuntu 24

pfuntner:
This is a simple change to `SystemdService.is_running()` to allow it to work for Ubuntu 24 to correct the symptoms in #775. 

On Ubuntu 24, `systemctl is-activate foobar` (assuming `foobar` is an unknown service) has an exit status of 4 but the `is_running()` method only expects exit status values of 0, 1, or 3.  My change just allows an exit status of 4 as well.  The method will return `False` anyway but it won't raise the `AssertionError`.

I was following `CONTRIBUTING.rst` and tried running `tox` from a virtual environment on my development machine but it failed and I'm not sure how to fix it.  Here's the output: [tox output.txt](https://github.com/user-attachments/files/16998670/tox.output.txt).

### DIFF
--- a/testinfra/modules/service.py
+++ b/testinfra/modules/service.py
@@ -181,7 +181,7 @@ class SystemdService(SysvService):
 
     @property
     def is_running(self):
-        out = self.run_expect([0, 1, 3], "systemctl is-active %s", self.name)
+        out = self.run_expect([0, 1, 3, 4], "systemctl is-active %s", self.name)
         if out.rc == 1:
             # Failed to connect to bus: No such file or directory
             return super().is_running


pfuntner:
It looks like the `tox` tests were run automatically on the Github side.  I didn't expect my tiny change would introduce a problem and I see it worked. 🥳

martinhoyer:
There is already https://github.com/pytest-dev/pytest-testinfra/pull/773, just no reviewers it seems :(

pfuntner:
> There is already #773, just no reviewers it seems :(

Oh, great.  Thanks, I'll close this one.

PR #782, opened 2024-11-04
Concatenate strings that are separated by a single space

CarstenGrohmann:
This is a cosmetic change to concatenate strings in a line that are only separated by a single space (`" "`).


CarstenGrohmann:
Not necessary anymore

PR #785, opened 2024-11-20
allow to limit ansible hosts with regex

marwin1991:
Hello, I have added simple change `urllib.parse.unquote(host)` for host variable to allow using regex.

I think is kinda not the best way, I was thinking about option `--ansible-limit` and in `ansble.py` we could change method `get_hosts` as follows:

```python
    @classmethod
    def get_hosts(cls, host: str, **kwargs: Any) -> list[str]:
        inventory = kwargs.get("ansible_inventory")
        hosts = AnsibleRunner.get_runner(inventory).get_hosts(host or "all")
        if kwargs.get("ansible_limit"):
            # Filter hosts based on the limit expression
            from ansible.parsing.dataloader import DataLoader
            from ansible.inventory.manager import InventoryManager

            loader = DataLoader()
            inventory_manager = InventoryManager(loader=loader, sources=inventory)
            limited_hosts = inventory_manager.get_hosts( pattern=kwargs.get("ansible_limit"))
            return [host for host in inventory if
                    host.name in [h.name for h in limited_hosts]]
        else:
            return hosts
```

what do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ flake8.report
 junit*.xml
 doc/build
 .cache
+.idea/
--- a/test/inventory.yml
+++ b/test/inventory.yml
@@ -1,0 +1,33 @@
+all:
+  hosts:
+    s1:
+      ansible_host: 192.168.1.1
+    s2:
+      ansible_host: 192.168.1.2
+    s3:
+      ansible_host: 192.168.1.3
+    s4:
+      ansible_host: 192.168.1.4
+    s5:
+      ansible_host: 192.168.1.5
+    s6:
+      ansible_host: 192.168.1.6
+    s7:
+      ansible_host: 192.168.1.7
+    s8:
+      ansible_host: 192.168.1.8
+    s9:
+      ansible_host: 192.168.1.9
+
+  children:
+    servers:
+      hosts:
+        s1:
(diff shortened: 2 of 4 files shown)


marwin1991:
@philpep ping :)

marwin1991:
If this change is okay I will create docs

testinfra/backend/ansible.py
```diff
@@ -87,4 +87,15 @@ def get_variables(self) -> dict[str, Any]:
     @classmethod
     def get_hosts(cls, host: str, **kwargs: Any) -> list[str]:
         inventory = kwargs.get("ansible_inventory")
-        return AnsibleRunner.get_runner(inventory).get_hosts(host or "all")
+        hosts = AnsibleRunner.get_runner(inventory).get_hosts(host or "all")
+        limit = kwargs.get("ansible_limit")
+        if limit:
+            # Filter hosts based on the limit expression
+            from ansible.parsing.dataloader import DataLoader
+            from ansible.inventory.manager import InventoryManager
```
osk-8, line 95: I think it will be clearer to bring the imports to the top of the file

marwin1991: @osk-8 done :)

Issue #786, opened 2024-11-29
when testing encodings for ansible - windows hosts error occurs [venv/lib/python3.12/site-packages/testinfra/backend/ansible.py:62: KeyError]


marwin1991:
![image](https://github.com/user-attachments/assets/fe9a52c8-9c78-4ca9-9bcd-9de834968720)

```
>       result = host.ansible(
            "win_shell",
            f"Get-Content -Path '{file_path}' | Out-String",
            check=False,
            become=True,
            become_user="vlan8user",
            become_method="runas",
        )

test_infra.py:20: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
venv/lib/python3.12/site-packages/testinfra/modules/ansible.py:43: in wrapper
    return func(self, *args, **kwargs)
venv/lib/python3.12/site-packages/testinfra/modules/ansible.py:110: in __call__
    result = self._host.backend.run_ansible(
venv/lib/python3.12/site-packages/testinfra/backend/ansible.py:72: in run_ansible
    result = self.ansible_runner.run_module(
venv/lib/python3.12/site-packages/testinfra/utils/ansible_runner.py:411: in run_module
    with open(fpath, "r", encoding=get_encoding()) as f:
venv/lib/python3.12/site-packages/testinfra/backend/ansible.py:70: in get_encoding
    return self.encoding
venv/lib/python3.12/site-packages/testinfra/backend/base.py:303: in encoding
    self._encoding = self.get_encoding()
venv/lib/python3.12/site-packages/testinfra/backend/base.py:283: in get_encoding
    cmd = self.run(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <testinfra.backend.ansible.AnsibleBackend object at 0x7f0793cdf9e0>
command = "python3 -c 'import locale;print(locale.getpreferredencoding())'"
args = ('python3',), kwargs = {'encoding': None}
[...]

PR #787, opened 2024-11-29
fix getting encoding for windows ansible hosts

marwin1991:
image


marwin1991:
This pull request fixes #786

PR #808, opened 2025-05-18
Fix quote logic and shell logic in Windows

### LINKED ISSUE #375: Incompatible arg quoting in windows
When running command with 1+ arguments against a docker host, the command automatically fails with `line 1: syntax error: unterminated quoted string`. The is because the `quote` method in `backend/base.py` is using `pipes.quote` (which is only compatible with posix systems) to generate quotes, and sometimes uses single quotes to enclose an arg. The default shell in windows is `cmd`, which does not accept single quoted args (only double quoted). On windows, a more accurate command would be something like this:

``` python
@staticmethod
def quote(command, *args):
    def anon_1(arg):
        if re.match(r'/("|\s|\')', arg) != None:
            return arg

        arg = re.sub('"', '\"', arg)
        return '"%s"' % (arg)

    if args:
        return command % tuple(anon_1(a) for a in args)
        # return command % tuple(pipes.quote(a) for a in args)
    return command
```

kayoub5:
fix #375 and #411

### DIFF
--- a/testinfra/backend/base.py
+++ b/testinfra/backend/base.py
@@ -14,6 +14,7 @@ import abc
 import dataclasses
 import locale
 import logging
+import platform
 import shlex
 import subprocess
 import urllib.parse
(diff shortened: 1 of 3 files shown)


kayoub5:
@philpep tdas3001 could you review please?

kayoub5:
@thomasleveil Thanks for the inspiration in https://github.com/pytest-dev/pytest-testinfra/issues/411#issuecomment-941057681

PR #816, opened 2026-02-28
Allow VSOCK as Ansible host

francoisfreitag:
From https://man.archlinux.org/man/vsock.7:
> The VSOCK address family facilitates communication between virtual machines and the host they are running on.

An example: https://wiki.archlinux.org/title/QEMU#Accessing_SSH_via_vsock

The separator between the "vsock" identifier and the CID can either be a forward slash (`/`) or a percent sign (`%`). `urllib.parse.urlparse` will mistake the forward slash for the beginning of the URL path, but a percent sign is accepted.

```python
>>> urlparse("ansible://vsock/555")
ParseResult(scheme='ansible', netloc='vsock', path='/555', params='', query='', fragment='')
>>> urlparse("ansible://vsock%555")
ParseResult(scheme='ansible', netloc='vsock%555', path='', params='', query='', fragment='')
```

### DIFF
--- a/testinfra/utils/ansible_runner.py
+++ b/testinfra/utils/ansible_runner.py
@@ -17,6 +17,7 @@ import ipaddress
 import json
 import os
 import tempfile
+import urllib.parse
 from collections.abc import Iterator
 from typing import Any, Callable, Optional, Union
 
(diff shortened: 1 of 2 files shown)


testinfra/utils/ansible_runner.py
```diff
@@ -214,7 +215,7 @@ def get_config(
     if version == 6:
         spec += "[" + testinfra_host + "]"
     else:
-        spec += testinfra_host
+        spec += urllib.parse.quote(testinfra_host)
```
francoisfreitag, line 218: This matches https://github.com/pytest-dev/pytest-testinfra/blob/c17794e9d74608c8adcb4c2a8d81fd62c29f53bc/testinfra/backend/base.py#L260

PR #817, opened 2026-05-01
Preserve explicit Paramiko username

### LINKED ISSUE #643: Paramiko backend overwrites username with ~/.ssh/config defaults
At the bottom of my ~/.ssh/config, I have this section because my local username doesn't match my most common remote username:

    Host *
    User kmccormick

When running pytest, any paramiko host always uses the kmccormick user, even though I specify vagrant. It's the same whether it is set on the command line or in conftest.py with get_host, e.g.:

    pytest --hosts "paramiko://vagrant@127.0.0.1:2222" 
    get_host("paramiko://vagrant@127.0.0.1:2222")

This breaks the convention with ssh and ssh_config where earlier options override later options; command line options should override anything in ssh_config. It looks like it's had the same behavior since ~/.ssh/config was added as a fallback in [5c3388c4](https://github.com/pytest-dev/pytest-testinfra/commit/5c3388c450d737749f499db6816cd82a42bd718c).

`_load_ssh_config()` should only set `cfg["username"] = value` if user isn't already set. That should probably also be extended out to other settings besides just username.

Herrtian:
Fixes #643.

When Paramiko loads ssh config defaults, a `Host *` user should not replace a username that was already supplied in the host spec, like `paramiko://vagrant@127.0.0.1:2222`.

This keeps the ssh config user fallback for host specs without a user, and adds a small regression test for both paths.

Checked with:

```
uv run --with paramiko python -
uv run --with ruff ruff check testinfra/backend/paramiko.py test/test_backends.py
uv run --with mypy --with types-paramiko --with paramiko --with pytest mypy testinfra/backend/paramiko.py
```

I also tried the focused pytest test on Windows, but this repo's test conftest hits Docker probing before collection and currently trips a Windows bytes-args issue in the local backend, so I verified the Paramiko config path directly instead.

### DIFF
--- a/test/test_backends.py
+++ b/test/test_backends.py
@@ -632,12 +632,12 @@ def test_paramiko_ssh_config_does_not_override_explicit_user():
 
     backend = ParamikoBackend("explicit@example.com")
     cfg = {"username": backend.host.user}
-    backend._load_ssh_config(object(), cfg, ssh_config)  # type: ignore[arg-type]
+    backend._load_ssh_config(paramiko.SSHClient(), cfg, ssh_config)
     assert cfg["username"] == "explicit"
 
     backend = ParamikoBackend("example.com")
     cfg = {"username": backend.host.user}
-    backend._load_ssh_config(object(), cfg, ssh_config)  # type: ignore[arg-type]
+    backend._load_ssh_config(paramiko.SSHClient(), cfg, ssh_config)
     assert cfg["username"] == "configuser"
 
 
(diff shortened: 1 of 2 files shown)


Herrtian:
Pushed 24ebb76 to fix the mypy failure from the now-unused type ignores. I checked `tox -e mypy`, `tox -e lint`, and `git diff --check` locally. The remaining py311 failure is in the existing Ansible encoding test, so I left that out of this Paramiko-focused change.

philpep (approved):

philpep:
Merged, thanks!

PR #819, opened 2026-05-17
Show GitHub stars in docs

### LINKED ISSUE #813: Display GitHub stars instead of watchers in the docs
Currently it looks like this:

Image

And at first glance, it is hard to understand that 67 is the number of watchers and not stars, and it gives an impression that the library is not as popular. I think instead of watchers it is better to show the number of stars.

Herrtian:
Closes #813.

Configures the Alabaster GitHub button to show stars instead of watchers.

Checks: sphinx-build -W -b html doc/source doc/build; pre-commit on doc/source/conf.py; git diff --check.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -125,6 +125,7 @@ html_theme_options = {
     "github_user": "pytest-dev",
     "github_repo": "pytest-testinfra",
     "github_button": True,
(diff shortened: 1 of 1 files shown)


philpep (approved):

philpep:
Merged, thanks!

PR #820, opened 2026-06-01
fix: handle ansible encoding errors and update rocky9 package assertion

shenxianpeng:
## Summary

Fixes CI failures in the `build (py311)` job:

### 1. Handle Ansible encoding errors gracefully (`testinfra/backend/ansible.py`)

When Ansible fails to run a command (e.g., due to UTF-8 surrogate encoding issues), it returns a response with only a `msg` key and no `stdout`/`stderr`. Previously this caused a `KeyError: 'stdout'`. Now the code:
- Checks for `module_stdout` (new ansible format)
- Falls back to `stdout` (backward compat)
- Falls back to `msg` for error responses, placed in `stderr`
- Uses `.get("rc", 1)` since `rc` may also be missing

### 2. Make encoding test resilient to Ansible version differences (`test/test_backends.py`)

The `test_encoding[ansible://debian_bookworm?force_ansible=True]` test now accepts either the old-style `ls` error message with surrogates (older ansible) or the new surrogate-rejection transport error from newer ansible versions. Also asserts `cmd.failed is True`.

### 3. Update Rocky Linux 9 package version expectation (`test/test_modules.py`)

The rocky linux 9 docker image now ships `openssh-server 9.9p1` instead of 8.x. Updated version prefix from `"8."` to `"9."`.

---

Fixes CI run: https://github.com/pytest-dev/pytest-testinfra/actions/runs/26749425227

### DIFF
--- a/test/test_backends.py
+++ b/test/test_backends.py
@@ -87,7 +87,8 @@ def test_encoding(host):
     elif host.backend.get_connection_type() == "ansible" and host.backend.force_ansible:
         # XXX: this encoding issue comes directly from ansible
         # not sure how to handle this...
-        assert (
+        assert cmd.failed
+        assert "surrogates not allowed" in cmd.stderr or (
             cmd.stderr
             == "ls: impossible d'acc\udce9der \udce0 '/é': Aucun fichier ou dossier de ce type"
         )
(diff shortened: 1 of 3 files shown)


philpep (approved):

philpep:
Merged, thanks!

PR #823, opened 2026-07-04
[NEW] Add Zoned Block Device (ZBD) support to blockdevice module

xsub:
### Motivation

With the growing adoption of SMR (Shingled Magnetic Recording) HDDs and NVMe ZNS (Zoned Namespace) drives, it's becoming more common to configure and test infrastructure that relies on Zoned Block Devices (ZBD). This PR extends the `blockdevice` module to allow querying ZBD-specific properties directly from Testinfra.

### Changes

This PR adds support for ZBD properties in `LinuxBlockDevice` by safely querying the Linux kernel's `sysfs` (`/sys/block/<dev>/queue/`). 

New properties added:
* `is_zoned` (bool)
* `zoned_type` (e.g., `"none"`, `"host-managed"`, `"host-aware"`)
* `zoned_chunk_sectors`
* `zoned_nr_zones`
* `zone_append_max_bytes`
* `max_open_zones`
* `max_active_zones`

**Implementation details:**
* **Safe attribute reading:** Instead of running shell commands (`cat`) and relying on exit codes or kernel version parsing, the code simply checks if the relevant `sysfs` file exists using `host.file().exists` and reads its content. 
* **Backward compatibility:** If a device is a standard block device (or if the kernel is too old to expose newer limits like `max_active_zones`), the module gracefully returns `None` for these specific attributes and `"none"` for `zoned_type`. It won't break existing tests.
* Used standard Python `@property` decorators for a clean API.

### Usage Example

```python
def test_nvme_zns_drive(host):
    drive = host.block_device("nvme1n1")
    
    assert drive.is_zoned is True
    assert drive.zoned_type == "host-managed"
    assert drive.max_open_zones == 14
```

### DIFF
--- a/testinfra/modules/blockdevice.py
+++ b/testinfra/modules/blockdevice.py
@@ -29,6 +29,7 @@ class BlockDevice(Module):
 
     def __init__(self, device):
         self.device = device
+        self._dev_name = device.replace("/dev/", "")
         super().__init__()
 
     @property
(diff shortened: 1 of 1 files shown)


xsub:
Hey Testinfra team!

Here's a small update for the blockdevice module to support Zoned Block Devices (ZBD). It adds a few properties like is_zoned, zoned_type, and various ZBD limits (max open zones, etc.).

Hope you find it useful.

Thanks,
xsub

